PR #1, opened 2023-07-21
Support executeWithSignerFee

hihiben:


packages/core/src/router.ts
```diff
@@ -39,3 +39,29 @@ export function newRouterExecuteTransactionRequest(
 
   return { to: getContractAddress(chainId, 'Router'), data, value };
 }
+
+export interface NewRouterExecuteWithSignerFeeTransactionOptions {
+  chainId: number;
+  routerBatchLogics: IParam.LogicBatchStruct;
+  signer: string;
+  signature: BytesLike;
+  tokensReturn?: string[];
```
Jeff-CCH, line 48:
```suggestion
  tokensReturn: string[];
```
Could remove `?` to skip the check for `undefined`

packages/core/src/router.ts
```diff
@@ -39,3 +39,29 @@ export function newRouterExecuteTransactionRequest(
 
   return { to: getContractAddress(chainId, 'Router'), data, value };
 }
+
+export interface NewRouterExecuteWithSignerFeeTransactionOptions {
```
Jeff-CCH, line 43: Does the name need to include the "New"? Would `RouterExecuteWithSignerFeeTransactionOptions` be enough?

packages/core/src/router.ts
```diff
@@ -39,3 +39,29 @@ export function newRouterExecuteTransactionRequest(
...
+  signature: BytesLike;
+  tokensReturn?: string[];
+  value?: BigNumberish;
+  referralCode?: number;
+}
+
+export function newRouterExecuteWithSignerFeeTransactionRequest(
+  options: NewRouterExecuteWithSignerFeeTransactionOptions
+): common.TransactionRequest {
+  const { chainId, routerBatchLogics, signer, signature, tokensReturn = [], value = 0, referralCode = 0 } = options;
```
Jeff-CCH, line 56: Why would we set `0` for `value` and `referralCode`? If that's the case, do users still need to provide these 2 params?

packages/core/src/router.ts
```diff
@@ -39,3 +39,29 @@ export function newRouterExecuteTransactionRequest(
...
+  const iface = Router__factory.createInterface();
+  const data = iface.encodeFunctionData('executeWithSignerFee', [
+    routerBatchLogics,
+    signer,
```
Jeff-CCH, line 60: Do we need to check the length of `signer` string before encoding it? (Applied to `tokensReturn` string array too)

Jeff-CCH: If we can use an `address` type here, the check would be unnecessary.

Jeff-CCH (commented):
Reviewed the implementation except the tests.

packages/core/src/router.ts
hihiben: I'm just following the original naming. @chouandy what do you think?

packages/core/src/router.ts
hihiben: I assume this is necessary to keep this to allowing `tokensReturn`, `value`, `referralCode` to be optional.

packages/core/src/router.ts
hihiben: It should be the way to assign default value since they are optional.

packages/core/src/router.ts
hihiben: I'm not sure if they are necessary. We can definitely check those including the length of signature. However the calldata will eventually be rejected when the call to contract is performed.

Jeff-CCH: Okay. I wondered if there is any chance of a hash collision when the length is unexpected. But for this case, developers can provide any data. Utilizing the length to make a hash collision would be too much work.

packages/core/src/router.ts
chouandy: My experience is function name + options, and I won't spend too much time thinking about what to omit.

chouandy (approved):
LGTM

packages/core/src/router.ts
Jeff-CCH: I see. Thanks for the explanation.

Jeff-CCH (approved):

Jeff-CCH:
LGTM

PR #5, opened 2023-10-06
Add Optimism, Base, Metis and Avalanche Info in Common Package

Jeff-CCH:
- [x] changeset
- [x] network info
- [x] router info

### DIFF
--- a/packages/common/src/networks/data.json
+++ b/packages/common/src/networks/data.json
@@ -135,7 +135,7 @@
       "symbol": "WETH",
       "name": "Wrapped Ether"
     },
-    "multicall2Address": "0x091e99cb1C49331a94dD62755D168E941AbD0693",
+    "multicall2Address": "0xcA11bde05977b3631167028862bE2a173976CA11",
     "multicall3Address": "0xcA11bde05977b3631167028862bE2a173976CA11"
   },
   {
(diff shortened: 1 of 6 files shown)


packages/common/src/networks/data.json
```diff
@@ -45,6 +68,76 @@
...
+      "name": "Ethereum"
+    },
+    "wrappedNativeToken": {
+      "chainId": 8453,
+      "address": "0x4200000000000000000000000000000000000006",
+      "decimals": 18,
+      "symbol": "WETH",
+      "name": "Wrapped Ether"
+    },
+    "multicall2Address": "0x091e99cb1C49331a94dD62755D168E941AbD0693",
```
chouandy: After checking, the contract interface at that address is not Multicall2. It is an optimized version modified by Uniswap. Since Multicall3 is backward compatible with Multicall2, we could use Multicall3 as a substitute for Multicall2 at the address.

Jeff-CCH: That's pretty weird since the Uniswap team put this address in the [Multicall2 address column](https://docs.uniswap.org/contracts/v3/reference/deployments). Will use the `multicall3Address` for the `multicall2Address`.

chouandy: It seems like Uniswap has already transitioned to using its own Multicall. I think I should find time to refine how we retrieve quotes from their interface.

Jeff-CCH: fixed

chouandy (approved):
LGTM

PR #11, opened 2023-11-20
feat: update router addresses for production use

zodahu:


Jeff-CCH (approved):
LGTM

packages/core/src/configs.ts
Jeff-CCH, line 1: Need to add Gnosis chain as well

chouandy (approved):
LGTM

PR #15, opened 2023-11-28
Add Lending SDK

LuPoYi:
Add Lending SDK


packages/lending/src/protocols/compound-v3/lending-protocol.ts
```diff
@@ -0,0 +1,357 @@
...
+  async getPortfolios(account: string) {
+    return await Promise.all(
+      Object.keys(marketMap[this.chainId]).map((marketId) => this.getPortfolio(account, marketId))
+    );
+  }
+
+  async newSupplyLogic(params: SupplyParams) {
+    const { cometAddress, baseToken } = getMarketBaseConfig(this.chainId, params.marketId);
+    const cToken = await this.getToken(cometAddress);
+    if (params.input.token.unwrapped.is(baseToken.unwrapped)) {
```
Jeff-CCH: Since the market config is maintained by this SDK (which is not the same as the compound-kit behavior), the condition here is different from the [condition in compound-kit](https://github.com/dinngo/compound-kit-api/blob/master/src/handlers/v1/markets/%5BchainId%5D/%5BmarketId%5D/zap-supply/index.ts#L116). Suggest to maintain the same behavior/condition.

packages/lending/src/protocols/compound-v3/lending-protocol.ts
```diff
@@ -0,0 +1,357 @@
...
+
+  newBorrowLogic = protocols.compoundv3.newBorrowLogic;
+
+  async newRepayLogic(params: RepayParams): Promise<any> {
+    const repayQuotation = await protocols.compoundv3.getRepayQuotation(this.chainId, {
+      tokenIn: params.input.token,
+      borrower: params.borrower,
+      marketId: params.marketId,
+    });
+    repayQuotation.input.amount = params.input.amount;
```
Jeff-CCH: reminder: if `params.input.amount > debt amount`, the extra amount will be supplied to compound v3.

packages/lending/src/protocols/compound-v3/lending-protocol.ts
```diff
@@ -0,0 +1,357 @@
...
+        balanceBps: common.BPS_BASE,
+      });
+    }
+  }
+
+  async newWithdrawLogic(params: WithdrawParams) {
+    const { cometAddress, baseToken } = getMarketBaseConfig(this.chainId, params.marketId);
+    const cToken = await this.getToken(cometAddress);
+
+    const realAmount = new common.TokenAmount(params.output.token, params.output.amount);
```
Jeff-CCH: reminder: if `realAmount > supplied amount` (when token = baseToken), the extra amount will be borrowed from compound v3.

packages/lending/src/protocols/compound-v3/lending-protocol.ts
```diff
@@ -0,0 +1,357 @@
...
+        balanceBps: common.BPS_BASE,
+      });
+      withdrawBaseLogic.fields.output.amount = realAmount.subWei(1).amount;
+      return withdrawBaseLogic;
+    } else {
+      return protocols.compoundv3.newWithdrawCollateralLogic({ output: params.output, marketId: params.marketId });
+    }
+  }
+
+  newBorrowLogic = protocols.compoundv3.newBorrowLogic;
```
Jeff-CCH: Do we need to check
* baseBorrowMin
* availableToBorrow

 If no precise error message shows up, it would be difficult for client to find out the root cause.

packages/lending/src/protocols/compound-v3/lending-protocol.ts
```diff
@@ -0,0 +1,357 @@
...
+    return portfolio;
+  }
+
+  async getPortfolios(account: string) {
+    return await Promise.all(
+      Object.keys(marketMap[this.chainId]).map((marketId) => this.getPortfolio(account, marketId))
+    );
+  }
+
+  async newSupplyLogic(params: SupplyParams) {
```
Jeff-CCH: Since we do not check if the user has borrowed position, is it legit that a user uses `newSupplyLogic` to do compound v3 repay?

packages/lending/src/protocols/aave-v2/lending-protocol.ts
```diff
@@ -0,0 +1,363 @@
...
+
+  toProtocolToken(underlyingToken: common.Token) {
+    return toAToken(this.chainId, underlyingToken);
+  }
+
+  isProtocolToken(token: common.Token) {
+    return isAToken(this.chainId, token);
+  }
+
+  async newSupplyLogic(params: SupplyParams) {
```
Jeff-CCH: Do we need to add the referral code?

packages/lending/src/protocols/aave-v2/lending-protocol.ts
```diff
@@ -0,0 +1,363 @@
...
+      input: {
+        token: toAToken(this.chainId, params.output.token),
+        amount: params.output.amount,
+      },
+      tokenOut: params.output.token,
+    });
+    return protocols.aavev2.newWithdrawLogic(withdrawQuotation);
+  }
+
+  newBorrowLogic = protocols.aavev2.newBorrowLogic;
```
Jeff-CCH: check available to borrow?

packages/lending/src/protocols/aave-v2/lending-protocol.ts
```diff
@@ -0,0 +1,363 @@
...
+
+  async newRepayLogic(params: RepayParams) {
+    if (!params.borrower || !params.interestRateMode) throw new Error('missing requied params');
```
Jeff-CCH: could add more detailed in error msg like missing borrower param for clients to understand easier.

packages/lending/src/protocols/aave-v3/lending-protocol.ts
```diff
@@ -0,0 +1,367 @@
...
+
+  toProtocolToken(underlyingToken: common.Token) {
+    return toAToken(this.chainId, underlyingToken);
+  }
+
+  isProtocolToken(token: common.Token) {
+    return isAToken(this.chainId, token);
+  }
+
+  async newSupplyLogic(params: SupplyParams) {
```
Jeff-CCH: add referral code?

packages/lending/src/protocols/openocean-v2/lending-swaper.ts
Jeff-CCH, line 1: file name: swaper -> swapper

packages/lending/src/protocols/openocean-v2/lending-swaper.ts
```diff
@@ -0,0 +1,38 @@
+import { NAME } from './configs';
+import { Swaper } from 'src/swaper';
+import * as api from '@protocolink/api';
```
Jeff-CCH: import name `api` is not the same as `protocols` in other files (eg. aave-v3:lending-protocol.ts)

packages/lending/src/protocols/openocean-v2/lending-swaper.ts
```diff
@@ -0,0 +1,38 @@
...
+  }
+
+  async tokens() {
+    if (!this._tokens) {
+      this._tokens = await api.protocols.openoceanv2.getSwapTokenTokenList(this.chainId);
+    }
+    return this._tokens!;
+  }
+
+  async quote(params: api.protocols.paraswapv5.SwapTokenParams): Promise<logics.openoceanv2.SwapTokenLogicFields> {
```
Jeff-CCH:
```suggestion
  async quote(params: api.protocols.openoceanv2.SwapTokenParams): Promise<logics.openoceanv2.SwapTokenLogicFields> {
```

packages/lending/src/protocols/openocean-v2/lending-swaper.ts
```diff
@@ -0,0 +1,38 @@
+import { NAME } from './configs';
+import { Swaper } from 'src/swaper';
+import * as api from '@protocolink/api';
+import * as common from '@protocolink/common';
+import * as logics from '@protocolink/logics';
+import { quote } from '@protocolink/api';
+
+export class LendingSwaper extends Swaper {
+  static readonly supportedChainIds = [1088];
```
Jeff-CCH: OpenOceanV2 is on Gnosis and Metis now. suggest get `supportedChainIds` from the SDK.

packages/lending/src/protocols/paraswap-v5/lending-swaper.ts
```diff
@@ -0,0 +1,38 @@
...
+  async quote(params: api.protocols.paraswapv5.SwapTokenParams): Promise<logics.paraswapv5.SwapTokenLogicFields> {
+    return await quote(this.chainId, logics.paraswapv5.SwapTokenLogic.rid, params);
+  }
+
+  newSwapTokenLogic(fields: api.protocols.paraswapv5.SwapTokenFields): api.protocols.paraswapv5.SwapTokenLogic {
+    return api.protocols.paraswapv5.newSwapTokenLogic(fields);
+  }
+
+  isExactIn() {
+    return true;
```
Jeff-CCH: Not sure the role this function plays in the whole architecture, but paraswap v5 can be exact in and exact out depending on the input param.

packages/lending/src/protocols/radiant-v2/lending-protocol.ts
```diff
@@ -0,0 +1,313 @@
...
+
+  toProtocolToken(underlyingToken: common.Token) {
+    return toRToken(this.chainId, underlyingToken);
+  }
+
+  isProtocolToken(token: common.Token) {
+    return isRToken(this.chainId, token);
+  }
+
+  async newSupplyLogic(params: SupplyParams) {
```
Jeff-CCH: Do we need to use the same verb as the protocol? (deposit for radiant v2)

packages/lending/test/transactions/zap-supply.ts
```diff
@@ -0,0 +1,173 @@
...
+          apporveTimes: 2,
+          recieves: [aaveV2.mainnetTokens.aWBTC],
+        },
+      },
+    ];
+
+    for (const [i, { skip, protocolId, marketId, params, expects }] of testCases.entries()) {
+      if (skip) continue;
+
+      it.only(`case ${i + 1} - ${protocolId}:${marketId}`, async function () {
```
Jeff-CCH:
```suggestion
      it(`case ${i + 1} - ${protocolId}:${marketId}`, async function () {
```

packages/lending/test/transactions/zap-supply.ts
```diff
@@ -0,0 +1,173 @@
...
+        params: {
+          srcToken: radiantV2.mainnetTokens.USDC,
+          srcAmount: '1',
+          destToken: radiantV2.mainnetTokens.WBTC,
+        },
+        expects: {
+          funds: [radiantV2.mainnetTokens.USDC],
+          balances: [radiantV2.mainnetTokens.rWBTC],
+          apporveTimes: 2,
+          recieves: [radiantV2.mainnetTokens.rWBTC],
```
Jeff-CCH:
```suggestion
          receives: [radiantV2.mainnetTokens.rWBTC],
```

packages/lending/test/transactions/zap-supply.ts
```diff
@@ -0,0 +1,173 @@
...
+  snapshotAndRevertEach();
+
+  context('Test Zap Supply', function () {
+    const testCases = [
+      {
+        skip: false,
+        protocolId: 'radiantv2',
+        marketId: 'mainnet',
+        params: {
+          srcToken: radiantV2.mainnetTokens.USDC,
```
Jeff-CCH:
```suggestion
          srcToken: mainnetTokens.USDC,
```
suggest: get common tokens from mainnetTokens and get protocol-specific tokens from protocols

packages/lending/test/transactions/zap-supply.ts
```diff
@@ -0,0 +1,173 @@
...
+        const transactionRequest = await zapDepositInfo.buildRouterTransactionRequest();
+        expect(transactionRequest).to.include.all.keys('to', 'data', 'value');
+
+        const tx = await user.sendTransaction(transactionRequest);
+        expect(tx).to.not.be.reverted;
+
+        for (const recv of expects.recieves) {
+          const balance = await getBalance(user.address, recv);
+
+          expect(balance.gt('0')).to.be.true;
```
Jeff-CCH: suggest: verify the balance difference of involved tokens to make sure the supply works as expected.

packages/lending/src/adapter.ts
```diff
@@ -0,0 +1,966 @@
...
+
+    portfolio = portfolio || (await protocol.getPortfolio(account, marketId));
+    const afterPortfolio = portfolio.clone();
+
+    // init with withdraw token amount
+    let outputTokenAmount = new common.TokenAmount(srcToken, srcAmount);
+
+    // ---------- withdraw ----------
+    const withdrawLogic = await protocol.newWithdrawLogic({
+      output: new common.TokenAmount(srcToken, srcAmount),
```
Jeff-CCH:
```suggestion
      output: outputTokenAmount
```

packages/lending/src/adapter.ts
```diff
@@ -0,0 +1,966 @@
...
+      marketId,
+    });
+    zapWithdrawlogics.push(withdrawLogic);
+
+    afterPortfolio.withdraw(srcToken, withdrawLogic.fields.output.amount);
+
+    // ---------- swap ----------
+    if (!srcToken.is(destToken)) {
```
Jeff-CCH: if srcToken is WETH and destToken is ETH, we can utilize Protocolink to do unwrap after to skip this swap. Same for the other way.

Jeff-CCH: please also check all the swaps to see if they need to be adjusted.

packages/lending/src/adapter.ts
```diff
@@ -0,0 +1,966 @@
...
+    protocolId: string,
+    marketId: string,
+    params: BaseParams,
+    account: string,
+    portfolio?: Portfolio
+  ): Promise<BaseFields> {
+    const { srcAmount } = params;
+    const srcToken = common.classifying(params.srcToken);
+    const destToken = common.classifying(params.destToken);
+    const wrappedSrcToken = wrapToken(this.chainId, srcToken);
```
Jeff-CCH: can use `params.srcToken.wrapped` to get wrapped token

packages/lending/src/adapter.ts
```diff
@@ -0,0 +1,966 @@
...
+    const wrappedDestToken = wrapToken(this.chainId, destToken);
+
+    const leverageLonglogics: api.Logic<any>[] = [];
+    const protocol = this.getProtocol(protocolId);
+
+    portfolio = portfolio || (await protocol.getPortfolio(account, marketId));
+    const afterPortfolio = portfolio.clone();
+
+    // ---------- Pre-calc quotation ----------
+    const swaper = this.findSwaper([wrappedDestToken, wrappedSrcToken]);
```
Jeff-CCH: How to exclude balancer if we borrow the flashloan from it?

packages/lending/src/adapter.ts
```diff
@@ -0,0 +1,966 @@
...
+      tokenIn: wrappedDestToken,
+      output: { token: wrappedSrcToken, amount: srcAmount },
+      slippage: defaultSlippage,
+    });
+    // convert swap type to exact in
+    swapQuotation = await swaper.quote({
+      input: swapQuotation.input,
+      tokenOut: wrappedSrcToken,
+      slippage: defaultSlippage,
+    });
```
Jeff-CCH: suggest moving this quote to swap section to avoid misunderstanding

packages/lending/src/adapter.ts
```diff
@@ -0,0 +1,966 @@
...
+    const swapTokenLogic = swaper.newSwapTokenLogic(swapQuotation);
+    leverageLonglogics.push(swapTokenLogic);
+
+    // ---------- supply ----------
+    const supplyLogic = await protocol.newSupplyLogic({ input: swapQuotation.output, marketId });
+    leverageLonglogics.push(supplyLogic);
+    afterPortfolio.supply(swapQuotation.output.token, swapQuotation.output.amount);
+
+    // ---------- return funds ----------
+    if (supplyLogic.fields.output) {
```
Jeff-CCH: Protocolink API would take care of the token return. could skip this part.

packages/lending/src/adapter.ts
```diff
@@ -0,0 +1,966 @@
...
+      buildRouterTransactionRequest,
+      logics: leverageLonglogics,
+    };
+  }
+
+  // tokenA: WBTC
+  // 1. flashloan WBTC
+  // 2. swap WBTC to USDC
+  // 3. deposit USDC, get aUSDC
+  // 4. return funds aWBTC to user
```
Jeff-CCH:
```suggestion
  // 4. return funds aUSDC to user
```

packages/lending/src/adapter.ts
```diff
@@ -0,0 +1,966 @@
...
+    //  get the quotation for how much collateral token is needed to exchange for the repay amount
+    let swapQuotation = await swaper.quote({
+      tokenIn: wrappedDestToken,
+      output: { token: wrappedSrcToken, amount: srcAmount },
+      slippage: defaultSlippage,
+    });
+    // convert swap type to exact in
+    swapQuotation = await swaper.quote({
+      input: swapQuotation.input,
+      tokenOut: wrappedSrcToken,
```
Jeff-CCH: seems the tokens used in input and tokenOut are reversed?

packages/lending/src/adapter.ts
```diff
@@ -0,0 +1,966 @@
...
+
+    // ---------- supply ----------
+    const supplyLogic = await protocol.newSupplyLogic({ input: swapQuotation.output, marketId });
+
+    collateralSwapLogics.push(supplyLogic);
+
+    afterPortfolio.supply(swapQuotation.output.token, swapQuotation.output.amount);
+
+    // compound v3 base token supplied not support collateral swap
+    if (protocolId === 'compoundv3' && supplyLogic.fields.output)
```
Jeff-CCH: Would it be better to check if supply token equals base token?
Also suggest doing this check at the beginning to save resources.

packages/lending/src/adapter.ts
```diff
@@ -0,0 +1,966 @@
...
+      this.permitType
+    );
+
+    const buildRouterTransactionRequest = (
+      args?: Omit<api.RouterData, 'chainId' | 'account' | 'logics'>,
+      apiKey?: string
+    ): Promise<common.TransactionRequest> =>
+      api.buildRouterTransactionRequest(
+        { ...args, chainId: this.chainId, account, logics: collateralSwapLogics },
+        apiKey ? { 'x-api-key': apiKey } : undefined
```
Jeff-CCH: Do other buildRouterTransactionRequest need to provide `apiKey`?
Also, where do we pass the `customFees` value?

packages/lending/src/protocols/openocean-v2/lending-swaper.ts
LuPoYi: fixed all `swaper` to `swapper`

packages/lending/src/protocols/openocean-v2/lending-swaper.ts
LuPoYi: Standardized all `@protocolink/api` import statements to use `import * as apisdk from '@protocolink/api'` for consistency.

packages/lending/src/protocols/paraswap-v5/lending-swaper.ts
LuPoYi: Removed.

packages/lending/src/adapter.ts
LuPoYi: adjusted all the swaps conditional statement 
from `if (!srcToken.is(destToken))`
to `if (!srcToken.unwrapped.is(destToken.unwrapped))`

packages/lending/src/protocols/openocean-v2/lending-swaper.ts
Jeff-CCH: Please also change the names in examples/

packages/lending/src/protocols/openocean-v2/lending-swaper.ts
LuPoYi: fix `logics.openoceanv2.SwapTokenLogic.supportedChainIds` to get supportedChainIds and put this to `config.ts`

```typescript
// config.ts
export const supportedChainIds = logics.openoceanv2.SwapTokenLogic.supportedChainIds;
``` 

```typescript
// lending-swapper.ts
export class LendingSwapper extends Swapper {
  static readonly supportedChainIds = supportedChainIds;
```

packages/lending/src/protocols/radiant-v2/lending-protocol.ts
LuPoYi: I think that in the SDK, we should maintain the consistency of the interface, unless there is a particular scenario that requires otherwise. Therefore, perhaps we can continue to use `newSupplyLogic`.

packages/lending/test/transactions/zap-supply.ts
LuPoYi: fixed

packages/lending/src/adapter.ts
LuPoYi: fixed

packages/lending/test/transactions/collateral-swap.ts
```diff
@@ -120,6 +125,10 @@ describe('Transaction: Collateral swap', function () {
         const tx = await user.sendTransaction(transactionRequest);
 
         expect(tx).to.not.be.reverted;
+
+        // after balance
+        const afterBalance = await getBalance(user.address, aaveV3.mainnetTokens.aEthWBTC);
+        console.log('after balance', JSON.stringify(afterBalance.amount.toString(), null, 2));
```
Jeff-CCH: suggest removing all the` console.log` after the debug is done.

packages/lending/src/protocols/openocean-v2/configs.ts
```diff
@@ -1 +1,5 @@
+import * as logics from '@protocolink/logics';
+
 export const NAME = 'openocean-v2';
```
Jeff-CCH: Would it be better if we use `ID` instead of `NAME`?

packages/lending/test/transactions/zap-supply.ts
LuPoYi: My first thought is that importing fewer packages would be better, and we don't need to consider whether it's a common token or protocol-specific token, so I think we'll stick with the current method for now.

packages/lending/src/adapter.ts
LuPoYi: The reason we keep swapQuotation here is that we should use the result `swapQuotation.input` as flashloan quote params
```typescript
    let swapQuotation = ...
    swapQuotation = ...

    // ---------- flashloan ----------
    const flashLoanAggregatorQuotation = await apisdk.protocols.utility.getFlashLoanAggregatorQuotation(this.chainId, {
      loans: [swapQuotation.input],
    });

```

packages/lending/src/adapter.ts
LuPoYi: The return funds here is that we should send aToken to user as collateral in the middle, and then they could borrow other token.

packages/lending/src/adapter.ts
LuPoYi: Add baseToken to `Portfolio` so that we can use it easily, and move this condition to save resources.

```typescript
// compound v3 base token supplied not support collateral swap
    if (protocolId === 'compound-v3' && portfolio.baseToken?.is(srcToken.unwrapped))
      throw new Error('Compound V3 does not support the base token for performing collateral swaps');
```

packages/lending/src/adapter.ts
LuPoYi: Added `apiKey` to `buildRouterTransactionRequest` for `debtSwap`  `leverageLong` `leverageShort` `deleverage`

`apiKey` is for `debtSwap` `collateralSwap` `leverageLong` `leverageShort` `deleverage`, other features don't need to pass `apiKey`.

we will pass the `customFees` on client side, like this
```typescript
const lendingInfo = await adapter.getLeverageShort(protocolId, marketId, params, account, portfolio);

const transactionRequest = await lendingInfo.buildRouterTransactionRequest(
  {
    permitSig,
    permitData,
    customFees
  },
   apiKey
);
```

packages/lending/src/protocols/openocean-v2/configs.ts
LuPoYi: Agree and changed.

packages/lending/src/adapter.ts
```diff
@@ -0,0 +1,988 @@
...
+      args?: Omit<apisdk.RouterData, 'chainId' | 'account' | 'logics'>
+    ): Promise<common.TransactionRequest> =>
+      apisdk.buildRouterTransactionRequest({ ...args, chainId: this.chainId, account, logics: zapBorrowlogics });
+
+    return {
+      fields: {
+        srcToken,
+        srcAmount,
+        destToken,
+
```
Jeff-CCH:
```suggestion
```

packages/lending/src/adapter.ts
weizard: In the leverage process, if a protocol similar to Aave is used at this step, collateral tokens should be returned to the user, enabling them to have sufficient collateral for the next step (borrowing).
If the protocol is Compound, the collateral will be accounted for directly on the user, and there is no need to return any tokens.

packages/lending/src/protocols/compound-v3/lending-protocol.ts
weizard: Updated, it's get the market config on chain for now.

packages/lending/src/protocols/compound-v3/lending-protocol.ts
weizard: added, make sense this is only for withdraw should not become to borrow.

packages/lending/src/protocols/compound-v3/lending-protocol.ts
weizard: Good point. Added.

packages/lending/src/protocols/compound-v3/lending-protocol.ts
weizard: Ok, updated.

packages/lending/src/adapter.ts
```diff
@@ -0,0 +1,991 @@
...
+        afterPortfolio,
+      },
+      estimateResult,
+      buildRouterTransactionRequest,
+      logics: leverageLonglogics,
+    };
+  }
+
+  // tokenA: WBTC
+  // 1. flashloan WBTC
```
Jeff-CCH: please use srcToken/destToken to present the token in the steps

packages/lending/src/adapter.ts
```diff
@@ -0,0 +1,991 @@
...
+    const swapper = this.findSwapper([wrappedDestToken, wrappedSrcToken]);
+    //  get the quotation for how much collateral token is needed to exchange for the repay amount
+    let swapQuotation = await swapper.quote({
+      tokenIn: wrappedDestToken,
+      output: { token: wrappedSrcToken, amount: srcAmount },
+      slippage: defaultSlippage,
+    });
+    // convert swap type to exact in
+    swapQuotation = await swapper.quote({
+      input: swapQuotation.input,
```
Jeff-CCH: suggest using srcToken/destToken instead of `xxx.input`

packages/lending/src/adapter.ts
```diff
@@ -0,0 +1,991 @@
...
+    const [flashLoanLoanLogic, flashLoanRepayLogic] = apisdk.protocols.utility.newFlashLoanAggregatorLogicPair(
+      flashLoanAggregatorQuotation.protocolId,
+      flashLoanAggregatorQuotation.loans.toArray()
+    );
+    collateralSwapLogics.push(flashLoanLoanLogic);
+
+    // ---------- swap ----------
+    const swapper = this.findSwapper([wrappedSrcToken, wrappedDestToken]);
+    const swapQuotation = await swapper.quote({
+      input: flashLoanTokenAmount,
```
Jeff-CCH: suggest using srcAmount

packages/lending/src/adapter.ts
```diff
@@ -0,0 +1,991 @@
...
+        { ...args, chainId: this.chainId, account, logics: collateralSwapLogics },
+        apiKey ? { 'x-api-key': apiKey } : undefined
+      );
+
+    return {
+      fields: {
+        srcToken: srcToken,
+        srcAmount: srcAmount,
+        destToken: destToken,
+        destAmount: withdrawLogic.fields.output.amount,
```
Jeff-CCH: should be `swapQuotation.output.amount`

packages/lending/src/adapter.ts
```diff
@@ -0,0 +1,991 @@
...
+    // ---------- Pre-calc quotation ----------
+    const swapper = this.findSwapper([wrappedSrcToken, wrappedDestToken]);
+    // get the quotation for how much dest token is needed to exchange for the src amount
+    let swapQuotation = await swapper.quote({
+      tokenIn: wrappedDestToken,
+      output: { token: wrappedSrcToken, amount: srcAmount },
+      slippage: defaultSlippage,
+    });
+    // convert swap type to exact in
+    swapQuotation = await swapper.quote({ input: swapQuotation.input, tokenOut: srcToken, slippage: defaultSlippage });
```
Jeff-CCH: Do we need another swap?

packages/lending/src/adapter.ts
```diff
@@ -0,0 +1,991 @@
...
+    const borrowTokenAmount = flashLoanAggregatorQuotation.repays.tokenAmountMap[wrappedDestToken.address];
+    const borrowLogic = protocol.newBorrowLogic({
+      output: borrowTokenAmount,
+      interestRateMode: defaultInterestRateMode,
+    });
+    debtSwapLogics.push(borrowLogic);
+    afterPortfolio.borrow(borrowTokenAmount.token, borrowTokenAmount.amount);
+
+    // ---------- flashloan repay ----------
+    debtSwapLogics.push(flashLoanRepayLogic);
```
Jeff-CCH: should consider flashloan fee

packages/lending/src/adapter.ts
```diff
@@ -0,0 +1,991 @@
...
+    ): Promise<common.TransactionRequest> =>
+      apisdk.buildRouterTransactionRequest(
+        { ...args, chainId: this.chainId, account, logics: leverageShortlogics },
+        apiKey ? { 'x-api-key': apiKey } : undefined
+      );
+
+    return {
+      fields: {
+        srcToken,
+        srcAmount,
```
Jeff-CCH: Do we need to return destAmount too?

packages/lending/src/protocols/openocean-v2/lending-swaper.ts
LuPoYi: fixed

packages/lending/src/protocols/radiant-v2/lending-protocol.ts
weizard: `LendingProtocol` implements the abstract `Protocol` to enable the adapter to utilize a unified interface for communication with each protocol. Therefore, the function name cannot be replaced with another name.

packages/lending/src/protocols/aave-v2/lending-protocol.ts
weizard: It appears that the referral code for flashloans is configured within the [API service](https://github.com/dinngo/protocolink-api/blob/master/src/controllers/aave-v3/flash-loan.ts#L8C7-L8C20). I have confirmed this with the PM. Following our discussion, it has been decided that all Aave referral codes will be configured in the API service.

packages/lending/src/protocols/aave-v3/lending-protocol.ts
weizard: It appears that the referral code for flashloans is configured within the [API service](https://github.com/dinngo/protocolink-api/blob/master/src/controllers/aave-v3/flash-loan.ts#L8C7-L8C20). I have confirmed this with the PM. Following our discussion, it has been decided that all Aave referral codes will be configured in the API service.

packages/lending/src/adapter.ts
hihiben: Why suggest using srcAmount? Isn't the swapping do the swap for the entire flashLoan amount?

packages/lending/src/protocols/compound-v3/lending-protocol.ts
weizard: As per our discussion, throwing errors would impact the current Furucombo interface. Therefore, the decision has been made to remove all error throws for now. We plan to have further discussions to design and implement a new error-handling mechanism at a later stage

packages/lending/src/adapter.ts
Jeff-CCH: found this does not include fees. resolve comment.

packages/lending/src/adapter.ts
Jeff-CCH: Yes. I was thinking that it's better to use all the variables in params (eg. srcToken, srcAmount) for reviewers to read the logic easily (it would be hard to check the value of flashLoanTokenAmount). But it's also right that swap amount comes from flashloan amount which comes from srcAmount. I guess this is a decision to make to see if we want to follow the flow of logic or to use limited variables for better understanding.

packages/lending/src/adapter.ts
```diff
@@ -0,0 +1,991 @@
...
+  // tokenA: WBTC
+  // 1. flashloan USDC
+  // 2. swap USDC to WBTC
+  // 3. repay WBTC
+  // 4. add fund rUSDC
+  // 5. withdraw USDC by rUSDC
+  // 6. flashloan repay WBTC
+  // * srcToken => repayToken
+  // * destToken => depositToken, collateralToken
+  async getDeleverage(
```
Jeff-CCH: suggest move on top of `getLeverageLong` to follow the order in `test/transactions`

packages/lending/src/adapter.ts
weizard: After discussion will add the dest token & amount in short & long

packages/lending/src/adapter.ts
weizard: Thanks, fixed it.

packages/lending/src/adapter.ts
```diff
@@ -214,12 +267,8 @@ export class Adapter extends common.Web3Toolkit {
         srcToken: srcToken,
         srcAmount: srcAmount,
         destToken: destToken,
-<<<<<<< Updated upstream
         // TODO: check which one is right
         destAmount: swapQuotation.output.amount, //withdrawLogic.fields.output.amount,
-=======
-        destAmount: swapQuotation.output.amount,
->>>>>>> Stashed changes
```
Jeff-CCH: I think we should keep the Stashed changes here. Also please help remove the comment at Ln. 270.

packages/lending/src/adapter.config.ts
```diff
@@ -0,0 +1,61 @@
...
+    chainId: common.ChainId.arbitrum,
+    primaryStablecoin: arbitrumTokens['USDC.e'],
+    secondaryStablecoin: arbitrumTokens.DAI,
+    primaryNonstablecoin: arbitrumTokens.ETH,
+  },
+  {
+    chainId: common.ChainId.optimism,
+    primaryStablecoin: optimismTokens.USDC,
+    secondaryStablecoin: optimismTokens.DAI,
+    primaryNonstablecoin: optimismTokens.ETH,
```
Jeff-CCH, line 34: Could I ask what are the meanings of `primaryStablecoin`, `secondaryStablecoin` and `primaryNonstablecoin`?

packages/lending/src/protocols/aave-v2/lending-protocol.ts
```diff
@@ -0,0 +1,359 @@
...
+      supplyAPY: string;
+      stableBorrowAPY: string;
+      variableBorrowAPY: string;
+    }
+  >;
+
+  async getReserveDataMap() {
+    if (!this._reserveDataMap) {
+      const calls: common.Multicall3.CallStruct[] = [];
+      for (const { asset } of this.reserves) {
```
Jeff-CCH, line 149: Would it be clearer if we use `reserve` instead of `asset`?

packages/lending/src/protocols/aave-v2/lending-protocol.ts
```diff
@@ -329,34 +328,32 @@ export class LendingProtocol extends Protocol {
...
+  async newSupplyLogic({ marketId, input }: TokenInFields) {
+    return apisdk.protocols.aavev2.newDepositLogic({
+      input,
+      output: new common.TokenAmount(this.toProtocolToken(marketId, input.token), input.amount),
     });
-    return apisdk.protocols.aavev2.newDepositLogic({ ...supplyQuotation, balanceBps: common.BPS_BASE });
   }
 
-  async newWithdrawLogic(params: WithdrawParams) {
-    const withdrawQuotation = await apisdk.protocols.aavev2.getWithdrawQuotation(this.chainId, {
```
Jeff-CCH: Is there any reason why we don't need `getWithdrawQuotation` anymore?

packages/lending/src/protocols/aave-v2/lending-protocol.ts
```diff
@@ -329,34 +328,32 @@ export class LendingProtocol extends Protocol {
...
+      output,
     });
-    return apisdk.protocols.aavev2.newWithdrawLogic(withdrawQuotation);
   }
 
-  newBorrowLogic = apisdk.protocols.aavev2.newBorrowLogic;
+  newBorrowLogic({ output }: TokenOutFields) {
+    return apisdk.protocols.aavev2.newBorrowLogic({
+      output,
+      interestRateMode: logics.aavev2.InterestRateMode.variable,
```
Jeff-CCH, line 355: I think it's nice to use the `interestRateMode` from `logics.aavev2`. However, the `defaultInterestRateMode` in `adapter.ts` should be adjusted as well.

packages/lending/src/protocols/aave-v2/lending-protocol.test.ts
```diff
@@ -0,0 +1,553 @@
...
+import * as common from '@protocolink/common';
+import { expect } from 'chai';
+
+describe('Test Aave V2 LendingProtocol', function () {
+  context('Test getPortfolio', function () {
+    const testCases = [
+      {
+        chainId: common.ChainId.mainnet,
+        account: '0xc94680947CF2114ec8eE43725898EAA7269a98c5',
+        blockTag: 18797586,
```
Jeff-CCH, line 11: Does hardhat create a new fork from the latest block and switch to `blockTag`? Or does hardhat create a fork directly from `blockTag`?

packages/lending/src/protocols/aave-v2/configs.ts
```diff
@@ -144,58 +115,42 @@ export const configs: Config[] = [
...
         asset: polygonTokens.WETH,
         aToken: polygonTokens.amWETH,
-        stableDebtTokenAddress: '0xc478cBbeB590C76b01ce658f8C4dda04f30e2C6f',
-        variableDebtTokenAddress: '0xeDe17e9d79fc6f9fF9250D9EEfbdB88Cc18038b5',
         used: { deposit: true, withdraw: true, borrow: true, repay: true, flashLoan: true },
       },
       {
         asset: polygonTokens.WMATIC,
         aToken: polygonTokens.amWMATIC,
-        stableDebtTokenAddress: '0xb9A6E29fB540C5F1243ef643EB39b0AcbC2e68E3',
```
Jeff-CCH: Is there any reason why we don't need `stableDebtTokenAddress` in the config?

packages/lending/src/protocols/aave-v2/configs.ts
```diff
@@ -16,7 +16,7 @@ export interface Reserve {
   };
 }
 
-type ContractName = 'LendingPool' | 'ProtocolDataProvider' | 'PriceOracle' | 'ETHPriceFeed';
```
Jeff-CCH: Is it because we can get lending pool address from `ProtocolDataProvider`?

packages/lending/src/protocol.ts
```diff
@@ -51,17 +51,15 @@ export abstract class Protocol extends common.Web3Toolkit {
 
   isAaveLike = false;
 
-  isUsingWrappedNativeToken = true;
-
   preferredFlashLoanerId?: string;
 
-  abstract newSupplyLogic(params: SupplyParams): Promise<SupplyLogic>;
+  abstract newSupplyLogic(params: any): api.Logic;
```
Jeff-CCH: Wondering why using `any` and `api.Logic` is a better option?

packages/lending/src/protocols/openocean-v2/lending-swapper.ts
```diff
@@ -2,32 +2,29 @@ import { ID, supportedChainIds } from './configs';
...
   async tokens() {
     if (!this._tokens) {
       this._tokens = await apisdk.protocols.openoceanv2.getSwapTokenTokenList(this.chainId);
     }
-    return this._tokens!;
+    return this._tokens;
   }
 
-  async quote(params: apisdk.protocols.openoceanv2.SwapTokenParams): Promise<logics.openoceanv2.SwapTokenLogicFields> {
-    return await apisdk.quote(this.chainId, logics.openoceanv2.SwapTokenLogic.rid, params);
```
Jeff-CCH: nice catch!

packages/lending/src/adapter.ts
```diff
@@ -0,0 +1,1051 @@
...
+    leverageShortlogics.push(swapTokenLogic);
+
+    // ---------- supply ----------
+    const supplyLogic = await protocol.newSupplyLogic({ input: swapQuotation.output, marketId, account });
+    leverageShortlogics.push(supplyLogic);
+
+    afterPortfolio.supply(swapQuotation.output.token, swapQuotation.output.amount);
+
+    // ---------- return funds ----------
+    if (supplyLogic.fields.output) {
```
Jeff-CCH: If compoundV3 does not support leverage short, we can remove this check.

packages/lending/src/adapter.ts
```diff
@@ -0,0 +1,1051 @@
...
+      interestRateMode: defaultInterestRateMode,
+      marketId,
+    });
+    deleveragelogics.push(repayLogic);
+    afterPortfolio.repay(swapQuotation.output.token, swapQuotation.output.amount);
+
+    // ---------- add funds ----------
+    if (protocolId !== 'compound-v3') {
+      const addLogic = apisdk.protocols.permit2.newPullTokenLogic({
+        input: swapQuotation.input,
```
Jeff-CCH: I think this should pull `aToken` instead of `destToken`?

packages/lending/src/adapter.ts
```diff
@@ -0,0 +1,1051 @@
...
+    marketId: string,
+    params: BaseParams,
+    account: string,
+    portfolio?: Portfolio
+  ): Promise<BaseFields> {
+    const { srcAmount } = params;
+    const srcToken = common.classifying(params.srcToken);
+    const destToken = common.classifying(params.destToken);
+    const wrappedSrcToken = srcToken.wrapped;
+    const wrappedDestToken = destToken.wrapped;
```
Jeff-CCH: Do we need `wrappedDestToken`? I guess we can use `destToken.wrapped` directly.

packages/lending/src/protocols/compound-v3/lending-protocol.ts
Jeff-CCH: please also remove the error in `getCollateralSwap`

packages/lending/test/transactions/zap-supply.ts
Jeff-CCH: The revised test cases use mainnetTokens for general tokens and use protocol.mainnetTokens for protocol specific tokens to reduce the amount of import and the length of single lines.

packages/lending/test/transactions/zap-supply.ts
Jeff-CCH: will verify the balance in the revised test cases

packages/lending/src/adapter.ts
```diff
@@ -0,0 +1,1051 @@
...
+    // get the quotation for how much dest token is needed to exchange for the src amount
+    let swapQuotation = await swapper.quote({
+      tokenIn: wrappedDestToken,
+      output: { token: wrappedSrcToken, amount: srcAmount },
+      slippage: defaultSlippage,
+    });
+    // convert swap type to exact in
+    swapQuotation = await swapper.quote({
+      input: swapQuotation.input,
+      tokenOut: srcToken,
```
Jeff-CCH: Would it be clearer to use `swapQuotation.output.token` in `tokenOut` to understand this is for exact in and also avoid bugs? (like this one should be `wrappedSrcToken` instead of `srcToken`).

packages/lending/src/protocol.type.ts
```diff
@@ -0,0 +1,60 @@
...
+  marketId: string;
+  input: common.TokenAmount;
+}
+
+export interface TokenOutFields {
+  marketId: string;
+  output: common.TokenAmount;
+}
+
+export type RepayFields = TokenInFields & { account: string };
```
Jeff-CCH: Wondering why we don't reuse the RepayFields in [core:logic-type.ts](https://github.com/dinngo/protocolink-js-sdk/blob/master/packages/core/src/logic-types.ts#L77C13-L77C24)

packages/lending/src/adapter.ts
```diff
@@ -0,0 +1,1051 @@
...
+      slippage: defaultSlippage,
```
Jeff-CCH: still trying to understand how this exact in works:

lets say srcAmount = 1000, srcToken:destToken = 1:1, defaultSlippage = 500 bps (5%)
at Ln.310, I will spend 1050 destTokens if 5% slippage is reached.
at 316, I will spend 1050 destTokens but only get 1050*0.95 = 997.5 srcTokens (if 5% slippage is reached).

Though it still works for the user, the repaid amount becomes 997.5 not 1000.
I'm wondering if we need to pay attention to this. Also, please let me know if any of the calculation is wrong. Thx.

packages/lending/src/protocols/compound-v3/lending-protocol.ts
```diff
@@ -449,107 +285,55 @@ export class LendingProtocol extends Protocol {
...
-  async newWithdrawLogic(params: WithdrawParams) {
-    const { cometAddress, baseToken } = getMarketBaseConfig(this.chainId, params.marketId);
-    const cToken = await this.getToken(cometAddress);
-
-    const userBalances = await this.getUserBalances(params.marketId, params.account);
+  async newWithdrawLogic({ marketId, output }: TokenOutFields) {
+    const { baseToken, cToken } = getMarketConfig(this.chainId, marketId);
 
-    const realAmount = new common.TokenAmount(params.output.token, params.output.amount);
-    realAmount.subWei(2);
```
Jeff-CCH: wondering why we removed this. This is related to 1 wei issue and will cause compound v3 withdraw fail

packages/lending/src/protocols/compound-v3/lending-protocol.ts
```diff
@@ -449,107 +285,55 @@ export class LendingProtocol extends Protocol {
...
-      const withdrawBaseLogic = await apisdk.protocols.compoundv3.newWithdrawBaseLogic({
-        ...withdrawBaseQuotation,
-        balanceBps: common.BPS_BASE,
+    if (output.token.wrapped.is(baseToken)) {
+      return apisdk.protocols.compoundv3.newWithdrawBaseLogic({
+        marketId,
+        input: new common.TokenAmount(cToken, output.amount),
+        output,
       });
-      withdrawBaseLogic.fields.output.amount = realAmount.subWei(1).amount;
```
Jeff-CCH: 1 wei issue

packages/lending/src/protocols/compound-v3/lending-protocol.ts
```diff
@@ -449,107 +285,55 @@ export class LendingProtocol extends Protocol {
...
-        input: {
-          token: cToken,
-          amount: realAmount.amount,
-        },
-        tokenOut: params.output.token,
-        marketId: params.marketId,
-      });
-      const withdrawBaseLogic = await apisdk.protocols.compoundv3.newWithdrawBaseLogic({
-        ...withdrawBaseQuotation,
-        balanceBps: common.BPS_BASE,
```
Jeff-CCH: remove BPS_BASE will cause withdraw failed

packages/lending/src/adapter.ts
```diff
@@ -0,0 +1,1051 @@
...
+    });
+
+    const [flashLoanLoanLogic, flashLoanRepayLogic] = apisdk.protocols.utility.newFlashLoanAggregatorLogicPair(
+      flashLoanAggregatorQuotation.protocolId,
+      flashLoanAggregatorQuotation.loans.toArray()
+    );
+    deleveragelogics.push(flashLoanLoanLogic);
+
+    // ---------- swap ----------
+    const swapTokenLogic = swapper.newSwapTokenLogic(swapQuotation);
```
Jeff-CCH: need to check if srcToken == destToken. plz check other swaps too.

ref
radiant-v2

packages/lending/src/protocol.portfolio.ts
```diff
@@ -0,0 +1,414 @@
...
+      amount = arg1;
+      supply.add(amount);
+    }
+    if (Number(amount) === 0) return;
+
+    const { price, apy, usageAsCollateralEnabled, ltv, liquidationThreshold } = supply;
+
+    const supplyUSD = new BigNumberJS(amount).times(price);
+    this.totalSupplyUSD = this.totalSupplyUSD.plus(supplyUSD);
+    this.totalCollateralUSD = this.totalCollateralUSD.plus(supplyUSD);
```
zodahu: Just discussed verbally, this line should be placed inside the `if (usageAsCollateralEnabled) {` block to ensure that only collateral can be summed, and should correspondingly modify the `calcNetAPY()` function.

c.c. @Jeff-CCH

Jeff-CCH: OK

packages/lending/src/protocols/compound-v3/lending-protocol.ts
zodahu: @chouandy is currently updating the encoding methods for each operation in `adapter.ts`. Issues related to `BalanceLink` and `wei` should ideally be handled in `adapter.ts`, rather than in individual lending-protocol files.

packages/lending/src/protocols/compound-v3/lending-protocol.ts
zodahu: @chouandy is currently updating the encoding methods for each operation in adapter.ts. Issues related to BalanceLink and wei should ideally be handled in adapter.ts, rather than in individual lending-protocol files.

packages/lending/src/protocols/compound-v3/lending-protocol.ts
zodahu: @chouandy is currently updating the encoding methods for each operation in adapter.ts. Issues related to BalanceLink and wei should ideally be handled in adapter.ts, rather than in individual lending-protocol files.

packages/lending/src/adapter.ts
Jeff-CCH: following the logic order would be easier to understand the flow. Also easier for debugging!

packages/lending/src/adapter.config.ts
LuPoYi: We currently use stablecoins as collateral tokens for leverage in both long and short positions.

When the leverage token is not USDC, we use `primaryStablecoin` as the collateral token. In all other cases, `secondaryStablecoin` is used as the collateral token.

`primaryNonstablecoin` as a non-stablecoin, common token.

packages/lending/src/protocol.portfolio.ts
zodahu: Fixed

packages/lending/src/protocol.ts
zodahu: @chouandy will fix after refactor

packages/lending/src/adapter.ts
zodahu: @chouandy will add this check

zodahu: This [commit](https://github.com/dinngo/protocolink-js-sdk/pull/15/commits/6b9ffc28025e318803e6e29b86ff0328d71f577a#diff-b79db2de06cbe1ac07ca0c06c1ff25c20b3b0334c36d073027b973c17a8d802a) has added this check for leverageLong, leverageShort, deleverage.

packages/lending/src/protocol.ts
zodahu: This [commit](https://github.com/dinngo/protocolink-js-sdk/pull/15/commits/7834a8ac89cbcb65ac7f6ff2a915e54f1d36e60c#diff-b79db2de06cbe1ac07ca0c06c1ff25c20b3b0334c36d073027b973c17a8d802a) has fixed.

PR #20, opened 2024-01-02
feat: borrow min and caps

chouandy:


packages/lending/src/adapter.ts
```diff
@@ -174,78 +174,86 @@ export class Adapter extends common.Web3Toolkit {
       const destCollateral = portfolio.findSupply(destToken);
 
       if (srcCollateral && destCollateral) {
-        // 1. validate src amount
-        if (new BigNumberJS(srcAmount).gt(srcCollateral.balance)) {
+        // 1. validate withdraw
+        if (!srcCollateral.validateWithdraw(srcAmount)) {
           output.error = { name: 'srcAmount', code: 'INSUFFICIENT_AMOUNT' };
         }
         output.afterPortfolio.withdraw(srcCollateral.token, srcAmount);
```
LuPoYi, line 181: Is this expected that if there are any `output.error`,  the `output.afterPortfolio` will still need to be calculated?

packages/lending/src/protocols/aave-v2/lending-protocol.ts
```diff
@@ -186,6 +194,8 @@ export class LendingProtocol extends Protocol {
             calculateCompoundedRate({ rate: variableBorrowRate.toString(), duration: SECONDS_PER_YEAR }),
             RAY_DECIMALS
           ),
+          totalSupply: common.toBigUnit(totalSupply, asset.decimals),
```
LuPoYi, line 197: `totalSupply` is from aToken, should we use `aToken.decimals` as `decimals` here? (the same as aavev3 and radiantv2)

packages/lending/src/adapter.ts
LuPoYi, line 1: https://github.com/chouandy/protocolink-js-sdk/blob/feature/borrow-min-and-caps/packages/lending/src/adapter.ts#L959

`const repayLogic = await protocol.newRepayLogic({ marketId, account, input: repayInput });`

Please remove the unnecessary `await`. Thanks.

packages/lending/src/adapter.ts
chouandy: The current process involves updating the after-portfolio based on the user's input. If adjustments are necessary, we should discuss with the product manager whether specifications need to be modified. cc @zodahu

packages/lending/src/protocols/aave-v2/lending-protocol.ts
chouandy: OK~

packages/lending/src/adapter.ts
chouandy: OK~

packages/lending/src/protocols/aave-v2/lending-protocol.ts
chouandy: As discussed, maintain the original underlying decimals

packages/lending/src/adapter.ts
chouandy: Fixed~

packages/lending/src/adapter.ts
```diff
@@ -174,78 +174,86 @@ export class Adapter extends common.Web3Toolkit {
...
-          // 5-1. if protocol is collateral tokenized
-          if (protocol.isAssetTokenized(marketId, withdrawOutput.token)) {
-            // 5-1-1. return dest protocol token to user
-            const returnFundsLogic = apisdk.protocols.utility.newSendTokenLogic({
-              input: supplyLogic.fields.output!,
-              recipient: account,
-            });
-            returnFundsLogic.fields.balanceBps = common.BPS_BASE;
-            output.logics.push(returnFundsLogic);
+          if (!output.error) {
```
Jeff-CCH, line 213: wondering if we can reduce the indents in this clause in two ways

```
1.  if(output.error) return ...
2. return after output.error = ....
```

Jeff-CCH:
Do we need to add unit tests to make sure users will get an error when borrow min and supply/borrow caps are reached?

packages/lending/src/adapter.ts
zodahu: The current implementation seems good as it allows users to preview their portfolio even if they don't have the funds. This enhances the user experience by enabling users to preview potential outcomes without needing to have the funds.

packages/lending/src/adapter.ts
chouandy: Sure, "early return" is one way of writing code, but it might result in many instances of `return output`. This approach might make the code feel less cohesive when read from top to bottom. What do other people think? @zodahu @LuPoYi

chouandy:
> Do we need to add unit tests to make sure users will get an error when borrow min and supply/borrow caps are reached?

I've considered it, but it's challenging to find the right user and market fit. If the PM thinks it's necessary, I'll spend time searching for it. cc @zodahu

packages/lending/src/adapter.ts
LuPoYi: Both approaches work for me. However, if I must choose, I have a slight preference for the "early return" approach in this situation.

zodahu: I don't have a strong opinion regarding this `adapter.ts` case, but we can maintain the current approach since this PR preserves the original style. If you have a strong preference, feel free to create a ticket with your reasons and assign it to me, or directly submit a PR for the elaboration. I'll then allocate resources for evaluation and potential refactoring.

packages/lending/src/adapter.ts
LuPoYi: OK, just to rephrase that: The current process is such that if there are any errors, it won't execute the code to build `destAmount` and `logics`. However, `afterPortfolio` will still be calculated regardless.

```typescript
export type OperationOutput = {
  destAmount: string;
  afterPortfolio: Portfolio;
  error?: OperationError;
  logics: apisdk.Logic[];
};
```

Jeff-CCH:
LGTM

LuPoYi:
LGTM

PR #24, opened 2024-01-08
feat(lending): add swapper quote error handling

chouandy:


packages/lending/src/adapter.ts
```diff
@@ -534,55 +556,55 @@ export class Adapter extends common.Web3Toolkit {
             // 1-5-1. the supply amount is the swap quotation output
             supplyInput = swapQuotation.output;
```
Jeff-CCH, line 562: Wondering if we need to add a try-catch for Ln.555

chouandy: Fixed~

Jeff-CCH:
LGTM

LuPoYi:
LGTM

LuPoYi (commented):
LGTM

PR #28, opened 2024-01-16
Support morpho blue on lending sdk

LuPoYi:


packages/lending/test/transactions/deleverage.test.ts
```diff
@@ -87,14 +103,29 @@ describe('Transaction: Deleverage', function () {
...
+      ) => {
         it(`case ${i + 1} - ${protocolId}:${marketId}`, async function () {
           user = await hre.ethers.getImpersonatedSigner(account);
           portfolio = await adapter.getPortfolio(user.address, protocolId, marketId);
 
           let initCollateralBalance, initBorrowBalance;
-          if (protocolId === 'compound-v3') {
-            const service = new logics.compoundv3.Service(chainId, hre.ethers.provider);
+
+          if (logicService) {
```
hihiben: I understand that this is a neat code that enables the test to differentiate [CompoundV3, Morpho] and other protocols, while providing `logics.{protocol}.Service` at the same time. However it's a little bit tricky since every protocol has `logics.{protocol}.Service`. Wondering that if there's a better way to do this.

packages/lending/test/transactions/leverage-long.test.ts
```diff
@@ -75,13 +78,29 @@ describe('Transaction: Leverage Long', function () {
...
       },
+      {
+        protocolId: 'morphoblue',
+        marketId: '0xb323495f7e4148be5643a4ea4a8221eef163e4bccfdedc2a6f4696baacbc86cc',
+        account: '0xa3C1C91403F0026b9dd086882aDbC8Cdbc3b3cfB',
+        srcToken: morphoblue.mainnetTokens.wstETH,
+        srcAmount: '0.001',
+        destToken: mainnetTokens.USDC,
+        srcAToken: undefined,
+        destDebtToken: undefined,
```
hihiben: Can be removed?

packages/lending/test/transactions/zap-repay.test.ts
```diff
@@ -81,69 +80,88 @@ describe('Transaction: Zap Repay', function () {
...
+          logicLength: 2,
+        },
+      },
+      {
+        protocolId: 'morphoblue',
+        marketId: '0xb323495f7e4148be5643a4ea4a8221eef163e4bccfdedc2a6f4696baacbc86cc',
+        account: '0xa3C1C91403F0026b9dd086882aDbC8Cdbc3b3cfB',
+        srcToken: mainnetTokens.USDC,
+        srcAmount: '0.1',
+        srcDebtToken: undefined,
```
hihiben: Can be removed?

packages/lending/package.json
```diff
@@ -24,10 +24,10 @@
...
   "dependencies": {
     "@aave/math-utils": "^1.21.0",
-    "@protocolink/api": "^1.0.4",
-    "@protocolink/common": "^0.3.5",
-    "@protocolink/core": "^0.4.5",
-    "@protocolink/logics": "^1.0.3",
+    "@protocolink/api": "^1.0.9",
+    "@protocolink/common": "^0.3.6",
+    "@protocolink/core": "^0.4.6",
+    "@protocolink/logics": "^1.0.8",
```
hihiben:
```suggestion
    "@protocolink/api": "^1.0.10",
    "@protocolink/common": "^0.3.6",
    "@protocolink/core": "^0.4.6",
    "@protocolink/logics": "^1.0.9",
```
These will be the latest version.

packages/lending/src/adapter.ts
```diff
@@ -573,6 +574,8 @@ export class Adapter extends common.Web3Toolkit {
         } catch (err) {
           output.error = err instanceof OperationError ? err : new OperationError('srcAmount', 'UNEXPECTED_ERROR');
         }
+      } else {
+        output.error = new OperationError('srcAmount', 'NOT_SUPPORTED_TOKEN');
```
Jeff-CCH:
```suggestion
        output.error = new OperationError('srcAmount', 'UNSUPPORTED_TOKEN');
```

packages/lending/src/adapter.ts
```diff
@@ -708,6 +711,8 @@ export class Adapter extends common.Web3Toolkit {
...
```
Jeff-CCH: Do we need to apply this to other scenarios (eg. zaps)?

packages/lending/src/protocols/morphoblue/adapter.test.ts
```diff
@@ -0,0 +1,666 @@
...
+import { Portfolio } from 'src/protocol.portfolio';
+import * as common from '@protocolink/common';
+import { expect } from 'chai';
+import { mainnetTokens } from './tokens';
+
+// Test information
+// account: 0xa3C1C91403F0026b9dd086882aDbC8Cdbc3b3cfB
+// pool: USDC(wstETH collateral)
+// balance: 0.001 wstETH (collateral) and 10 USDC (Loan)
+// blockTag: 18982784
```
Jeff-CCH: Ln.8 to Ln.12 can be removed

packages/lending/src/protocols/morphoblue/adapter.test.ts
```diff
@@ -0,0 +1,666 @@
...
+
+describe('Test Adapter for Morpho Blue', function () {
+  const chainId = common.ChainId.mainnet;
+  const blockTag = 18982784;
```
Jeff-CCH, line 10: I think `blockTag` is not required since we need to call paraswap and paraswap only works for the latest few blocks

packages/lending/src/protocols/morphoblue/adapter.test.ts
```diff
@@ -0,0 +1,666 @@
...
+
+  const adapter = new Adapter(chainId);
+
+  const marketId = '0xb323495f7e4148be5643a4ea4a8221eef163e4bccfdedc2a6f4696baacbc86cc';
+  const protocol = new LendingProtocol(chainId);
+  protocol.setBlockTag(blockTag);
+
+  // Morpho blue - only one collateral token and one debt token
+  // leverage only have one scenario - Long collateral token by using debt token
+  context('Test leverageLong', function () {
```
Jeff-CCH, line 20: Please review all this tests and make sure they are passed in the local environment and [the CI environment](https://github.com/dinngo/protocolink-js-sdk/actions/runs/7538789469/job/20519992315?pr=28)

packages/lending/src/protocols/morphoblue/lending-protocol.ts
```diff
@@ -0,0 +1,229 @@
...
+  override canDeleverage() {
+    return true;
+  }
+
+  toUnderlyingToken(marketId: string) {
+    const { collateralToken } = getMarket(this.chainId, marketId);
+    return collateralToken;
+  }
+
+  toProtocolToken(marketId: string) {
```
Jeff-CCH: I thinks this function is for protocols which have their own tokens like aToken and cToken. For Morpho, it does not have any token so this function should return `undefined`?

packages/lending/src/protocols/morphoblue/lending-protocol.ts
```diff
@@ -0,0 +1,229 @@
...
+
+  override canLeverage(_marketId: string) {
+    return true;
+  }
+
+  override canDeleverage() {
+    return true;
+  }
+
+  toUnderlyingToken(marketId: string) {
```
Jeff-CCH: same as the comment of Ln. 207

packages/lending/src/protocols/morphoblue/lending-protocol.ts
```diff
@@ -0,0 +1,229 @@
...
+        target: loanTokenPriceFeedAddress,
+        callData: this.priceFeedIface.encodeFunctionData('latestAnswer'),
+      },
+      {
+        target: oracle,
+        callData: this.oracleIface.encodeFunctionData('price'),
+      },
+    ];
+
+    const { returnData } = await this.multicall3.callStatic.aggregate(calls, { blockTag: this.blockTag });
```
Jeff-CCH, line 98: What kind of scenario we will need `blockTag`?

packages/lending/src/protocols/morphoblue/lending-protocol.ts
```diff
@@ -0,0 +1,229 @@
...
+    ).toString();
+
+    const supplies: SupplyObject[] = [
+      {
+        token: collateralToken,
+        price: collateralTokenPrice,
+        balance: supplyBalance,
+        apy: '0',
+        usageAsCollateralEnabled: true,
+        ltv: ltv,
```
Jeff-CCH:
```suggestion
        ltv,
```

packages/lending/src/protocols/morphoblue/lending-protocol.ts
```diff
@@ -0,0 +1,229 @@
...
+
+    const loan = totalBorrowAssets.mul(borrowShares).div(totalBorrowShares);
+
+    const supplyBalance = common.toBigUnit(collateral, collateralToken.decimals);
+    const borrowBalance = common.toBigUnit(loan.toString(), loanToken.decimals);
+    const totalSupply = common.toBigUnit(totalSupplyAssets, collateralToken.decimals);
+    const totalBorrow = common.toBigUnit(totalBorrowAssets, loanToken.decimals);
+
+    const loanTokenPrice = common.toBigUnit(_loanTokenPrice, 8);
+    const collateralTokenPrice = common.toBigUnit(
```
Jeff-CCH, line 122: I'm wondering if we need to calculate the collateral-usd price. For Morpho, the oracle in the market should be the only one to get the price no matter what. This makes me think about what should users do when the collatera-usd price is not the same as the chainlink price.

I guess two things here

1. What is `collateralTokenPrice` used for?
2. Should  `collateralTokenPrice` be quoted in `USD` or be quoted in `loanToken`?

packages/lending/src/protocols/morphoblue/lending-protocol.ts
```diff
@@ -0,0 +1,229 @@
...
+      _loanTokenPrice.mul(normalizedCollateralLoanPrice).div(BigNumber.from(10).pow(normalizedCollateralLoanDecimals)),
```
Jeff-CCH, line 123: Where did this formula come from?

packages/lending/src/protocols/morphoblue/lending-protocol.ts
```diff
@@ -0,0 +1,229 @@
...
+      totalSupplyAssets,
+      totalSupplyShares,
+      totalBorrowAssets,
+      totalBorrowShares,
+      lastUpdate,
+      fee,
+    };
+    const borrowRateView = await irmContract.borrowRateView(marketParams, market, { blockTag: this.blockTag });
+    const borrowApy = (
+      Math.exp(new BigNumberJS(borrowRateView.toString()).div(1e18).times(SECONDS_PER_YEAR).toNumber()) - 1
```
Jeff-CCH: could we put the calculation in `utils`?

packages/lending/src/protocols/morphoblue/lending-protocol.ts
```diff
@@ -0,0 +1,229 @@
...
+  get irmIface() {
+    if (!this._irmIface) {
+      this._irmIface = Irm__factory.createInterface();
+    }
+    return this._irmIface;
+  }
+
+  getMarketName(id: string) {
+    const { loanToken, collateralToken } = getMarket(this.chainId, id);
+    return `${DISPLAY_NAME} ${loanToken.symbol}(${collateralToken.symbol} collateral)`;
```
Jeff-CCH, line 73: When would we use this function? For me, the market info is still confusing if I want to know more than the tokens.

packages/lending/src/protocols/morphoblue/lending-protocol.ts
hihiben: CompoundV3 has a function `getAPYs` implemented.

packages/lending/src/protocols/morphoblue/lending-protocol.ts
hihiben: Having the same question here. Do we have a definition for underlying token and protocol token? In CompoundV3 `toUnderlyingToken` returns base token, which is loan token.

packages/lending/src/adapter.ts
LuPoYi: Good point, I will implement this for all scenarios as well.

packages/lending/src/protocols/morphoblue/lending-protocol.ts
LuPoYi: I think `blockTag` is only used for testing. In our tests, like `lending-protocol.test.ts`, we write down the expected output to ensure that the results is the same. If we don't set the `blockTag`, the results will change at any time.

packages/lending/src/protocols/morphoblue/lending-protocol.ts
LuPoYi: Implemented `getBorrowAPY` with the `irmContract.borrowRateView` contract call included.

packages/lending/src/protocols/morphoblue/lending-protocol.ts
LuPoYi: The functions `toUnderlyingToken` and `toProtocolToken` are used for handling supply tokens and their own tokens. For example, passing USDC would return aUSDC, and the other way around. This is primarily designed for Aave-like protocols. In the case of Compoundv3, the base token also serves as the supply token. 

For Morpho blue, I think using `undefined` here makes sense.

packages/lending/test/transactions/leverage-long.test.ts
LuPoYi: removed

packages/lending/test/transactions/zap-repay.test.ts
LuPoYi: removed

packages/lending/package.json
LuPoYi: updated

packages/lending/src/adapter.ts
LuPoYi: thanks and fixed

packages/lending/src/adapter.ts
LuPoYi: Implemented this for all scenarios when the supply or borrow token isn't found

packages/lending/src/protocols/morphoblue/adapter.test.ts
LuPoYi: removed

packages/lending/src/protocols/morphoblue/adapter.test.ts
LuPoYi: All tests have passed successfully in both the local and CI environments

packages/lending/src/protocols/morphoblue/lending-protocol.ts
LuPoYi: We use this function `getMarketName` on the frontend to display the title of the market, such as `Aave V3` `Compound V3 USDC.e` `Morpho Blue USDC(ETH collateral)`

packages/lending/src/protocols/morphoblue/lending-protocol.ts
LuPoYi: fixed

packages/lending/src/protocols/morphoblue/lending-protocol.ts
LuPoYi: The formula is from [morpho blue oracles](https://docs.morpho.org/contracts/morpho-blue/reference/oracles/#price). I've also included the link in the code comments above.

packages/lending/src/protocols/morphoblue/lending-protocol.ts
LuPoYi: 1. We display both the collateral and loan token prices on the lending dashboard to easily show the value of the tokens.
2. Currently, the prices are quoted in USD to align with other protocols. At the moment, I'm not sure of the use case for quoting prices in loanToken.

packages/lending/src/protocols/morphoblue/lending-protocol.ts
LuPoYi: Morpho blue operates with a single loan and single collateral market, where the `LLTV` is equal to both the `ltv`(max LTV) and the `liquidationThreshold`.


Also, a note on naming: the `ltv` property here should ideally be renamed to `maxLTV`. We could consider addressing this after the current MR is completed.

packages/lending/test/transactions/deleverage.test.ts
LuPoYi: I'm still considering how to improve this approach. For now, should I revert to using `if CompoundV3 else if Morpho blue` to handle it? I am open to this.

packages/lending/src/protocols/morphoblue/lending-protocol.ts
hihiben: In that case, it might be necessary to show some info about the marketId, or it is acceptable to have the same name?

packages/lending/test/transactions/deleverage.test.ts
Jeff-CCH: I guess we can stay with `if-else` first. If the checking condition grows too big someday, we can then figure out a way to improve it.

packages/lending/src/adapter.ts
```diff
@@ -319,6 +319,8 @@ export class Adapter extends common.Web3Toolkit {
         } catch (err) {
           output.error = err instanceof OperationError ? err : new OperationError('srcAmount', 'UNEXPECTED_ERROR');
         }
+      } else {
+        output.error = new OperationError(srcCollateral ? 'destAmount' : 'srcAmount', 'UNSUPPORTED_TOKEN');
```
Jeff-CCH, line 323: I thought `srcCollateral` responds to `srcAmount`?

packages/lending/src/protocol.portfolio.ts
```diff
@@ -284,6 +284,10 @@ export class Portfolio {
     return abbreviateUSD(this.totalBorrowUSD.toFixed());
   }
 
+  get formattedAvailableBorrowCapacityUSD() {
```
Jeff-CCH, line 287: I didn't find any function which would call this one. Is this for lending frontend to use?

packages/lending/src/protocols/morphoblue/lending-protocol.ts
```diff
@@ -0,0 +1,234 @@
...
+      totalSupplyAssets,
+      totalSupplyShares,
+      totalBorrowAssets,
+      totalBorrowShares,
+      lastUpdate,
+      fee,
+    };
+    const borrowApy = await this.getBorrowAPY(marketId, market);
+
+    const maxLtv = common.toBigUnit(lltv, 18);
```
Jeff-CCH, line 136: I'm a little bit confusing that why we need this transform considering Ln146 & Ln147 need string as well?

packages/lending/src/protocols/morphoblue/lending-protocol.ts
```diff
@@ -191,6 +176,8 @@ export class LendingProtocol extends Protocol {
     return false;
   }
 
+  canLeverageShort = false;
```
Jeff-CCH, line 177: Wondering why Morpho cannot leverage short?

packages/lending/src/protocols/morphoblue/lending-protocol.ts
```diff
@@ -199,14 +186,12 @@ export class LendingProtocol extends Protocol {
     return true;
   }
 
-  toUnderlyingToken(marketId: string) {
-    const { collateralToken } = getMarket(this.chainId, marketId);
-    return collateralToken;
+  toUnderlyingToken() {
+    return undefined;
```
Jeff-CCH: I guess we need to figure out the definition of underlying token for morpho. It could be loan token if we follow the spirit of compound v3

packages/lending/src/protocols/morphoblue/lending-protocol.ts
```diff
@@ -217,6 +202,27 @@ export class LendingProtocol extends Protocol {
...
+  async getBorrowAPY(marketId: string, market: MarketStruct) {
+    const { loanToken, collateralToken, oracle, irm, lltv } = getMarket(this.chainId, marketId);
+
+    const marketParams: MarketParamsStruct = {
+      loanToken: loanToken.address,
+      collateralToken: collateralToken.address,
+      oracle,
+      irm,
+      lltv,
+    };
```
Jeff-CCH: I think we can skip Ln.208 - Ln.214 and use the prams in Ln.218 directly.

packages/lending/src/protocols/morphoblue/lending-protocol.ts
```diff
@@ -217,6 +202,27 @@ export class LendingProtocol extends Protocol {
...
+
+    const irmContract = Irm__factory.connect(irm, this.provider);
+
+    const borrowRateView = await irmContract.borrowRateView(marketParams, market, { blockTag: this.blockTag });
+    const borrowApy = (
+      Math.exp(new BigNumberJS(borrowRateView.toString()).div(1e18).times(SECONDS_PER_YEAR).toNumber()) - 1
```
Jeff-CCH, line 208: Wondering why we need to transform `borrowRateView` to `BigNumberJS` when `borrowRateView` is already a `BigNumber`?

packages/lending/test/transactions/deleverage.test.ts
```diff
@@ -177,6 +177,7 @@ describe('Transaction: Deleverage', function () {
           // 5-1. debt grows when the block of getting api data is different from the block of executing tx
           const [minRepay] = utils.bpsBound(repayAmount.amount, 100);
           const minRepayAmount = repayAmount.clone().set(minRepay);
+
```
Jeff-CCH: unnecessary blank line

packages/lending/test/transactions/leverage-long.test.ts
```diff
@@ -130,7 +130,8 @@ describe('Transaction: Leverage Long', function () {
 
           if (logicService) {
             service = new logicService(chainId, hre.ethers.provider);
-            initSupplyBalance = await service.getCollateralBalance(marketId, user.address, leverageAmount.token);
+            initSupplyBalance = await service.getCollateralBalance(marketId, user.address, srcToken);
+
```
Jeff-CCH, line 120: the blank line could be removed

packages/lending/src/protocol.ts
```diff
@@ -53,9 +53,9 @@ export abstract class Protocol extends common.Web3Toolkit {
     return true;
   }
 
-  abstract toUnderlyingToken(marketId: string, protocolToken: common.Token): common.Token;
+  abstract toUnderlyingToken(marketId: string, protocolToken: common.Token): common.Token | undefined;
 
-  abstract toProtocolToken(marketId: string, underlyingToken: common.Token): common.Token;
+  abstract toProtocolToken(marketId: string, underlyingToken: common.Token): common.Token | undefined;
```
Jeff-CCH: Is it better to remove `abstract` and return `undefined` in `protocol.ts` directly?

packages/lending/src/protocol.utils.ts
```diff
@@ -84,9 +84,14 @@ export function abbreviateUSD(num: number | string, digits = 2) {
 
 export function toLessThanString(amount: string | BigNumberJS, decimals: number) {
   amount = new BigNumberJS(amount);
-  const minAmount = new BigNumberJS(1).shiftedBy(-decimals);
 
-  return amount.gt(minAmount) ? amount.toFixed() : `<${minAmount.toFixed()}`;
+  if (amount.isZero()) {
+    return '0';
+  } else {
```
Jeff-CCH: guess we don't need this else and can save an indent here

packages/lending/src/protocols/morphoblue/lending-protocol.ts
Jeff-CCH: Would it be weird if the same token has difference price on our UI due to different oracle? If we need to display the collateral price in usd, would it be better to get the price from chainlink?

packages/lending/test/hooks.ts
```diff
@@ -1,11 +1,12 @@
...
-import { aavev2, aavev3, compoundv3, radiantv2 } from 'src/protocols';
+import { aavev2, aavev3, compoundv3, morphoblue, radiantv2 } from 'src/protocols';
 import { paraswapv5 } from 'src/swappers';
 
 export async function setup() {
   Adapter.registerProtocol(aavev2.LendingProtocol);
   Adapter.registerProtocol(aavev3.LendingProtocol);
   Adapter.registerProtocol(compoundv3.LendingProtocol);
   Adapter.registerProtocol(radiantv2.LendingProtocol);
+  Adapter.registerProtocol(morphoblue.LendingProtocol);
```
Jeff-CCH, line 9: should place after compound v3 which follows the same order of the protocols folder

packages/lending/test/transactions/deleverage.test.ts
```diff
@@ -87,14 +103,29 @@ describe('Transaction: Deleverage', function () {
...
+          srcToken,
+          srcAmount,
+          srcDebtToken,
+          destToken,
+          destAToken,
+          logicService,
+          expects,
+        },
+        i
+      ) => {
```
Jeff-CCH: could we put all this in one line as before?

packages/lending/test/transactions/deleverage.test.ts
```diff
@@ -125,8 +156,7 @@ describe('Transaction: Deleverage', function () {
           await expect(user.sendTransaction(transactionRequest)).to.not.be.reverted;
 
           let collateralBalance, deleverageWithdrawAmount, borrowBalance;
-          if (protocolId === 'compound-v3') {
-            const service = new logics.compoundv3.Service(chainId, hre.ethers.provider);
+          if (service) {
```
Jeff-CCH, line 148: same here

packages/lending/src/protocol.portfolio.ts
LuPoYi: Yes, the frontend will get the `portfolio` instance, and then call the function they need.

packages/lending/src/adapter.ts
LuPoYi:
```
if (srcCollateral && destCollateral) {
  // skip
} else {
  output.error = new OperationError(srcCollateral ? 'destAmount' : 'srcAmount', 'UNSUPPORTED_TOKEN');
}
```

We checks if both `srcCollateral` and `destCollateral` are present. If either or both are missing, it enters the else block. Here, a ternary operator determines whether the error is with `srcCollateral` or `destCollateral`

packages/lending/src/protocols/morphoblue/lending-protocol.ts
LuPoYi: `lltv` value is like `860000000000000000` (from the `config.ts`)
`ltv` or `liquidationThreshold` value is `0.86` (for client side)

packages/lending/src/protocols/morphoblue/lending-protocol.ts
LuPoYi: Morpho only one collateral token and one debt token, so just one scenario: long collateral token by using debt token.

packages/lending/src/protocols/morphoblue/lending-protocol.ts
```diff
@@ -0,0 +1,236 @@
...
+  }
+
+  override isAssetTokenized(_marketId: string, _assetToken: common.Token) {
+    return false;
+  }
+
+  async getBorrowAPY(marketId: string, market: MarketStruct) {
+    const { loanToken, collateralToken, oracle, irm, lltv } = getMarket(this.chainId, marketId);
+
+    const irmContract = Irm__factory.connect(irm, this.provider);
```
hihiben, line 194: Should we apply the same mechanism like `get morpho()`?

packages/lending/src/protocols/morphoblue/lending-protocol.ts
hihiben: I thought long means that you're borrowing the token you want to long against. On the other hand, short means that you're borrowing the token you want to short.

Jeff-CCH (approved):
LGTM

hihiben (approved):
LGTM

PR #35, opened 2024-02-01
Add open position and close position

Jeff-CCH:
- [x] add open position and close position
- [x] unit test
- [x] e2e test
- [x] yarn changeset

### DIFF
--- a/packages/lending/src/adapter.ts
+++ b/packages/lending/src/adapter.ts
@@ -200,7 +200,7 @@ export class Adapter extends common.Web3Toolkit {
    * @param {common.Token} input.zapToken - Zap token: zap the token to the collateral.
    * @param {string} input.zapAmount - The amount of the zap token.
    * @param {common.Token} input.collateralToken - Collateral token: the collateral to be supplied.
-   * @param {string} input.collateralAmount - The supply amount of the collateral.
+   * @param {string} input.collateralAmount - The leverage amount of the collateral.
    * @param {common.Token} input.debtToken - Debt token: the debt to be borrowed.
    * @param {number} [input.slippage=defaultSlippage] - The slippage tolerance. Optional.
    * @returns {Promise<OperationOutput>} The result including the destination amount,
@@ -232,7 +232,7 @@ export class Adapter extends common.Web3Toolkit {
 
       if (srcCollateral && destBorrow) {
         try {
-          // 1. ---------- zap swap ----------
+          // 1. ---------- swap ----------
           const supplyInput = new common.TokenAmount(collateralToken.wrapped, '0');
           if (zapToken.is(collateralToken.wrapped)) {
             // 1-1-1. exclude native zapToken
@@ -354,6 +354,7 @@ export class Adapter extends common.Web3Toolkit {
             supplyInput.set(swapQuotation.output);
           }
 
+          // 2. ---------- leverage ----------
           if (debtAmount && Number(debtAmount) > 0) {
             output.afterPortfolio.supply(supplyInput.token, supplyInput.amount);
 
@@ -367,7 +368,7 @@ export class Adapter extends common.Web3Toolkit {
 
             if (leverageOutput.error) throw leverageOutput.error;
 
-            // 3-1. find supply logic and add zap supply amount
+            // 2-1-1. find supply logic and add zap supply amount
             const logicIndex = collateralToken.wrapped.is(debtToken.wrapped) ? 1 : 2;
             const leverageSupplyAmount = leverageOutput.logics[logicIndex].fields.input;
             const totalSupplyAmount = leverageSupplyAmount.add(supplyInput).amount;
--- a/packages/lending/src/adapter.ts
+++ b/packages/lending/src/adapter.ts
@@ -200,7 +200,7 @@ export class Adapter extends common.Web3Toolkit {
    * @param {common.Token} input.zapToken - Zap token: zap the token to the collateral.
    * @param {string} input.zapAmount - The amount of the zap token.
    * @param {common.Token} input.collateralToken - Collateral token: the collateral to be supplied.
-   * @param {string} input.collateralAmount - The total supplied collateral amount.
+   * @param {string} input.collateralAmount - The supply amount of the collateral.
    * @param {common.Token} input.debtToken - Debt token: the debt to be borrowed.
    * @param {number} [input.slippage=defaultSlippage] - The slippage tolerance. Optional.
    * @returns {Promise<OperationOutput>} The result including the destination amount,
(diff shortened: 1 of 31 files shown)


packages/lending/src/adapter.ts
```diff
@@ -185,6 +185,299 @@ export class Adapter extends common.Web3Toolkit {
...
+        }
+        const swapTokenLogic = swapper.newSwapTokenLogic(swapQuotation);
+        output.logics.push(swapTokenLogic);
+        // 2-1. the supply amount is the swap quotation output
+        supplyInput = swapQuotation.output;
+        output.afterPortfolio.supply(supplyInput.token, supplyInput.amount);
+      } else if (zapToken.wrapped.is(collateralToken.wrapped)) {
+        supplyInput = new common.TokenAmount(zapToken, zapAmount);
+        output.afterPortfolio.supply(supplyInput.token, supplyInput.amount);
+      }
```
LuPoYi: Users calling the `openByCollateral` function should ensure they provide the correct `zapToken` and `zapAmount` . I believe it's unnecessary for us to handle cases where `zapAmount` equals "0", they can use `Leverage` instead.

If it is deemed necessary, we can include a check in this section by verifying that `zapAmount !== '0'` before proceeding further
```
if (zapAmount !== '0'){
  if (zapToken.wrapped.is(collateralToken.wrapped)) {
    // ...
  }else{
    // zapSupply
  }
 ```
 
   if not, we can throw the error on the validate-params part

File: packages/lending/src/adapter.ts [hunk 2]
```diff
@@ -185,6 +185,299 @@ export class Adapter extends common.Web3Toolkit {
...
+
+      const destBorrow = portfolio.findBorrow(debtToken);
+      if (!destBorrow) throw new OperationError('debtToken', 'UNSUPPORTED_TOKEN');
+
+      const supplyAmountDelta = new common.TokenAmount(collateralToken.wrapped, collateralAmount).sub(
+        srcCollateral.balance
+      );
+      if (!srcCollateral.validateSupplyCap(supplyAmountDelta.amount)) {
+        throw new OperationError('collateralAmount', 'SUPPLY_CAP_EXCEEDED');
+      }
```
LuPoYi: Since this part involves validating parameters, should we also check if the amount is positive and throw error if it does?
```
if (Number(collateralAmount) < 0) throw new OperationError('collateralAmount', 'INVALID_AMOUNT');
```

packages/lending/src/adapter.ts
```diff
@@ -465,7 +758,7 @@ export class Adapter extends common.Web3Toolkit {
    * 5. borrow destToken
    * 6. flashloan repay destToken
    */
-  async leverageLong({
+  async leverageByCollateral({
```
zodahu, line 797: Could you help search and remove `long` and `short` terms and fix comments?

packages/lending/src/adapter.ts
zodahu: I believe that both `collateralAmount` and `zapAmount` being set to 0 should be considered valid inputs and should return the default `OperationOutput`. When users enter 0 in the FE, we should not trigger an error popup. Please refer to `collateralSwap`, which checks `srcAmount` at the beginning.

packages/lending/src/adapter.ts
```diff
@@ -185,6 +185,299 @@ export class Adapter extends common.Web3Toolkit {
...
+          });
+        } catch {
+          throw new OperationError('zapAmount', 'NO_ROUTE_FOUND_OR_PRICE_IMPACT_TOO_HIGH');
+        }
+        const swapTokenLogic = swapper.newSwapTokenLogic(swapQuotation);
+        output.logics.push(swapTokenLogic);
+        // 2-1. the supply amount is the swap quotation output
+        supplyInput = swapQuotation.output;
+        output.afterPortfolio.supply(supplyInput.token, supplyInput.amount);
+      } else if (zapToken.wrapped.is(collateralToken.wrapped)) {
```
zodahu: Could you please review whether we can replace `else if` with `else` since `zapAmount` is not being checked in this context? The absence of an `else` statement, when not dealing with error handling, seems like a potential error imo.

Moreover
1. then we can move `output.afterPortfolio.supply(supplyInput.token, supplyInput.amount);` outside of `if-else` block.
2. let `if-else` condition have no `!`

packages/lending/src/adapter.ts
```diff
@@ -185,6 +185,299 @@ export class Adapter extends common.Web3Toolkit {
...
+        leverageOutput.logics[2].fields.input.amount = supplyInput.add(
+          leverageOutput.logics[2].fields.input.amount
+        ).amount;
+        leverageOutput.logics[2].fields.balanceBps = common.BPS_BASE;
+      } else {
+        leverageOutput.logics[1].fields.input.amount = supplyInput.add(
+          leverageOutput.logics[1].fields.input.amount
+        ).amount;
+        leverageOutput.logics[1].fields.balanceBps = common.BPS_BASE;
+      }
```
zodahu: Can we swap the `if` and `else` blocks to eliminate `!` and keep consistency with other functions?

And we can simply this code by set index such as 
```
const logicIndex = collateralToken.wrapped.is(debtToken.wrapped) ? 1 : 2;

File: packages/lending/src/adapter.ts [hunk 6]
```diff
@@ -185,6 +185,299 @@ export class Adapter extends common.Web3Toolkit {
     return this.protocolMap[id];
   }
 
+  async openByCollateral(
```
zodahu: Add JSDoc for 3 new functions

packages/lending/src/adapter.ts
```diff
@@ -185,6 +185,299 @@ export class Adapter extends common.Web3Toolkit {
...
+        ).amount;
+        leverageOutput.logics[2].fields.balanceBps = common.BPS_BASE;
+      } else {
+        leverageOutput.logics[1].fields.input.amount = supplyInput.add(
+          leverageOutput.logics[1].fields.input.amount
+        ).amount;
+        leverageOutput.logics[1].fields.balanceBps = common.BPS_BASE;
+      }
+      output.logics.push(...leverageOutput.logics);
+      output.afterPortfolio = leverageOutput.afterPortfolio;
```
zodahu, line 269: Is it expected to override the whole `output.afterPortfolio` when step 2. have set `output.afterPortfolio.supply(...)`?

packages/lending/src/adapter.ts
```diff
@@ -185,6 +185,299 @@ export class Adapter extends common.Web3Toolkit {
...
+      output.destAmount = output.afterPortfolio.findBorrow(debtToken)!.balance;
```
zodahu: I think it's reasonable to return `leverageOutput.afterPortfolio.destAmount` as it represents the debt diff amount.

packages/lending/src/adapter.ts
zodahu: As both `leverageByCollateral` and `leverageByDebt` have handle supply and borrow cap checks, can we remove those cap valications from open functions?

packages/lending/src/adapter.ts
```diff
@@ -185,6 +185,299 @@ export class Adapter extends common.Web3Toolkit {
...
+            destToken: withdrawalToken.wrapped,
+            slippage,
+            isRepayAll: true,
+          });
+
+          if (zapRepayOutput.error) throw zapRepayOutput.error;
+
+          output.logics.push(...zapRepayOutput.logics);
+          flashLoanRepay.add(zapRepayOutput.destAmount);
+          output.afterPortfolio = zapRepayOutput.afterPortfolio;
```
zodahu, line 412: Is it expected to override `output.afterPortfolio` in loop?

packages/lending/src/adapter.ts
```diff
@@ -185,6 +185,299 @@ export class Adapter extends common.Web3Toolkit {
...
+      // 3. leverage by debt
+      const initDebtAmount = new common.TokenAmount(destBorrow.token, destBorrow.balance);
+      const leverageAmount = new common.TokenAmount(destBorrow.token, debtAmount).sub(initDebtAmount).amount;
+      const leverageOutput = await this.leverageByDebt({
+        account,
+        portfolio,
+        srcToken: debtToken.wrapped,
+        srcAmount: leverageAmount,
+        destToken: supplyInput.token,
+      });
```
zodahu:
```suggestion
      const leverageOutput = await this.leverageByDebt({
        account,
        portfolio: output.afterPortfolio,
        srcToken: debtToken.wrapped,
        srcAmount: leverageAmount,
        destToken: collateralToken, // Be consistent with `openByCollateral`
      });
```

Using `collateralToken` be consistent with `openByCollateral`

packages/lending/src/adapter.ts
Jeff-CCH: Since the cap validation in `openByDebt` is after `leverageByDebt`, agree that we can remove this one.

However, for `openByCollateral`, the validation is before zap-swap & `leverageByCollateral`, which means we can save API calls if we do the validation earlier. Does that make sense?

packages/lending/src/adapter.ts
Jeff-CCH: IMO, `zapAmount` could be 0 but `collateralAmount` should be gte the user's supplied balance.

packages/lending/src/adapter.ts
Jeff-CCH: I was struggling with whether we need to return the diff debt amount or the total debt amount here. Considering that I'm a user, I would prefer to get the total debt amount to calculate the health factor by using this value and `collateralAmount`. What does @LuPoYi think?

packages/lending/src/adapter.ts
Jeff-CCH: After adding a few new validations in the new commit, the validateSupplyCaps will be removed since they are redundant.

packages/lending/src/adapter.ts
```diff
@@ -213,58 +236,55 @@ export class Adapter extends common.Web3Toolkit {
...
-        let swapQuotation: SwapperQuoteFields;
-        try {
-          swapQuotation = await swapper.quote({
-            input: new common.TokenAmount(zapToken, zapAmount),
-            tokenOut: collateralToken.wrapped,
-            slippage,
-          });
-        } catch {
-          throw new OperationError('zapAmount', 'NO_ROUTE_FOUND_OR_PRICE_IMPACT_TOO_HIGH');
+      // 1-1. ---------- zap swap ----------
```
zodahu: Can re-serialize the numbers.

packages/lending/src/adapter.ts
LuPoYi: prefer to use the debt diff amount to check what changes if I do this transaction

packages/lending/src/adapter.ts
```diff
@@ -185,6 +191,335 @@ export class Adapter extends common.Web3Toolkit {
...
+
+      // 2. leverage by collateral
+      const leverageAmount = supplyAmountDelta.sub(supplyInput).amount;
+      const leverageOutput = await this.leverageByCollateral({
+        account,
+        portfolio: output.afterPortfolio,
+        srcToken: collateralToken.wrapped,
+        srcAmount: leverageAmount,
+        destToken: debtToken,
+      });
```
zodahu: Considering that the `leverageByCollateral`, `leverageByDebt`, and `zapWithdraw` functions are capable of processing both native and wrapped tokens, is it feasible to keep non-wrapped tokens in open/close functions?

packages/lending/src/adapter.ts
zodahu: Based on our recent verbal discussion, since the leverage part does not manage existing positions, we can directly use the `destAmount` returned by `leverageOutput`.

LuPoYi:
Please assist in adding `open` and `close` to the type `Operation` in protocol.type.ts.

packages/lending/src/adapter.ts
Jeff-CCH: since `leverageXXX` does not handle native token and we can not let `zapWithdraw` handle native token (cause withdraw is followed by flashloan repay), open & close need to handle native token. But `debtToken` needs to be changed to `debtToken.wrapped` here.

zodahu:
LGTM

LuPoYi:
LGTM

PR #36, opened 2024-02-14
build positions in e2e tests

Jeff-CCH:
- [x] build positions in e2e tests
- [x] yarn changeset

### DIFF
--- a/packages/lending/test/transactions/collateral-swap.test.ts
+++ b/packages/lending/test/transactions/collateral-swap.test.ts
@@ -1,15 +1,12 @@
 import { Adapter } from 'src/adapter';
 import { SignerWithAddress } from '@nomiclabs/hardhat-ethers/signers';
 import { TokenAmount } from '@protocolink/common';
-import * as aaveV2 from 'src/protocols/aave-v2/tokens';
-import * as aaveV3 from 'src/protocols/aave-v3/tokens';
 import * as apisdk from '@protocolink/api';
 import { claimToken, mainnetTokens, snapshotAndRevertEach } from '@protocolink/test-helpers';
 import * as common from '@protocolink/common';
 import { expect } from 'chai';
 import hre from 'hardhat';
 import * as logics from '@protocolink/logics';
-import * as radiantV2 from 'src/protocols/radiant-v2/tokens';
 import * as spark from 'src/protocols/spark/tokens';
 import * as utils from 'test/utils';
 
@@ -35,10 +32,8 @@ describe('Transaction: Collateral swap', function () {
         protocolId: 'aave-v2',
         marketId: 'mainnet',
         srcToken: mainnetTokens.WETH,
-        srcAmount: initSupplyAmount,
-        srcAccountingToken: aaveV2.mainnetTokens.aWETH,
+        srcAmount: '1',
         destToken: mainnetTokens.WBTC,
-        destAccountingToken: aaveV2.mainnetTokens.aWBTC,
         expects: { logicLength: 7 },
       },
       {
@@ -46,9 +41,7 @@ describe('Transaction: Collateral swap', function () {
         marketId: 'mainnet',
         srcToken: mainnetTokens.WETH,
         srcAmount: '1',
-        srcAccountingToken: radiantV2.mainnetTokens.rWETH,
         destToken: mainnetTokens.WBTC,
-        destAccountingToken: radiantV2.mainnetTokens.rWBTC,
         expects: { logicLength: 7 },
       },
       {
@@ -56,9 +49,7 @@ describe('Transaction: Collateral swap', function () {
         marketId: 'mainnet',
         srcToken: mainnetTokens.WETH,
         srcAmount: '1',
-        srcAccountingToken: aaveV3.mainnetTokens.aEthWETH,
         destToken: mainnetTokens.WBTC,
-        destAccountingToken: aaveV3.mainnetTokens.aEthWBTC,
         expects: { logicLength: 7 },
       },
       {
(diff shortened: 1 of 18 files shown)


packages/lending/test/utils/index.ts
```diff
@@ -1,5 +1,6 @@
 export * from './utils';
-export * from './compound-v3';
-export * from './aave-v3';
-export * from './aave-v2';
-export * from './radiant-v2';
+// export * from './compound-v3';
+// export * from './aave-v3';
+// export * from './aave-v2';
+// export * from './radiant-v2';
```
zodahu: Unused lines

packages/lending/test/transactions/collateral-swap.test.ts
```diff
@@ -32,132 +34,130 @@ describe('Transaction: Collateral swap', function () {
...
-            destBalance = await service.getCollateralBalance(marketId, user.address, destToken);
-          } else {
-            srcBalance = await getBalance(user.address, srcAToken!);
-            destBalance = await getBalance(user.address, destAToken!);
-          }
-
           // 4. user's src token balance will decrease.
-          expect(srcBalance.gte(initSrcBalance.clone().sub(srcAmount))).to.be.true;
+          const initSrcBalance = new TokenAmount(srcToken, initSupplyAmount);
+          const srcBalance = await utils.getCollateralBalance(chainId, protocolId, marketId, user, srcAccountingToken);
```
zodahu: Can we use each protocol's `toAToken`, `toRToken` in `utils.getCollateralBalance` to eliminate the need for parameters like `srcAccountingToken`, `destAccountingToken` in `testCases`?

packages/lending/test/transactions/deleverage.test.ts
```diff
@@ -16,168 +15,196 @@ import * as utils from 'test/utils';
...
-        srcDebtToken: '0x72E95b8931767C79bA4EeE721354d6E99a61D004', // variableDebtEthUSDC
-        destToken: mainnetTokens.WBTC,
-        destAToken: aaveV3.mainnetTokens.aEthWBTC,
-        expects: {
-          logicLength: 6,
-        },
+        srcAmount: '100',
+        variableDebtToken: '0x72E95b8931767C79bA4EeE721354d6E99a61D004', // variableDebtEthUSDC
+        destToken: mainnetTokens.WETH,
+        protocolToken: aaveV3.mainnetTokens.aEthWETH,
```
zodahu: `variableDebtToken` -> `srcDebtToken`: keeps the same naming as debt-swap, zaps
`protocolToken` -> When `utils.getCollateralBalance` modified as the above comment, we don't need this

packages/lending/test/transactions/deleverage.test.ts
```diff
@@ -16,168 +15,196 @@ import * as utils from 'test/utils';
...
+          const [minRepayAmount, maxRepayAmount] = utils.bpsBound(repayAmount.amount, slippage);
           expect(borrowDifference.gte(minRepayAmount)).to.be.true;
-          expect(borrowDifference.lte(repayAmount)).to.be.true;
+          expect(borrowDifference.lte(maxRepayAmount)).to.be.true;
         });
       }
     );
   });
+
+  context('Test Deleverage Approve', function () {
```
zodahu: The naming convention for the tests involving both `compound` and `morpho` is unclear, as neither requires `permit2` approve. If the intention is to identify cases where the permit sign flow is necessary, could we introduce a boolean flag within the `testCases`? This approach would keep 1 `context` within this file.

packages/lending/test/utils/morphoblue.ts
```diff
@@ -0,0 +1,85 @@
...
+      user.address
+    )
+  ).to.not.be.reverted;
+}
+
+export async function getCollateralBalance(
+  chainId: number,
+  marketId: string,
+  user: SignerWithAddress,
+  collateral: common.Token
```
zodahu: Can we use `service.getCollateralToken` instead of having parameter?

packages/lending/test/transactions/collateral-swap.test.ts
```diff
@@ -32,132 +34,130 @@ describe('Transaction: Collateral swap', function () {
...
-        destAToken: aaveV2.mainnetTokens.aWETH,
-        expects: {
-          logicLength: 7,
-        },
+        srcToken: mainnetTokens.WETH,
+        srcAmount: initSupplyAmount,
+        srcAccountingToken: aaveV2.mainnetTokens.aWETH,
+        destToken: mainnetTokens.WBTC,
+        destAccountingToken: aaveV2.mainnetTokens.aWBTC,
+        expects: { logicLength: 7 },
```
LuPoYi: initSupplyAmount is 2, so the final logics will have `Add Funds: 2 aWETH`. I think the user's address doesn't have enough aWETH to pass the test.

LuPoYi: It's okay, my mistake, just see the part about the deposit

packages/lending/test/transactions/deleverage.test.ts
Jeff-CCH: the test name will be refined as `Test XXX build positions` & `Test XXX on-chain positions` if that's more clear.

For the flag, I prefer to keep 2 contexts. If we merge them into one, we will have input used only by one side (eg. account). We will also have if-else conditions to decide the test flow, which would complicate the test.

packages/lending/test/transactions/collateral-swap.test.ts
Jeff-CCH: The `initSupplyAmount` is 2 and the `collateralSwapAmount` is 2 which seems no problem (and the test was passed). But for consistency, I will change the amount to 1 as other test cases.

zodahu:
LGTM

LuPoYi:
LGTM

PR #42, opened 2024-03-18
Add ZeroEx V4

Jeff-CCH:
- [x] add zeroex v4
- [x] yarn changeset

### DIFF
--- a/packages/api/src/protocols/zeroex-v4/swap-token.ts
+++ b/packages/api/src/protocols/zeroex-v4/swap-token.ts
@@ -1,0 +1,25 @@
+import { Logic } from 'src/types';
+import * as common from '@protocolink/common';
+import { getProtocolTokenList, quote } from 'src/api';
+import * as logics from '@protocolink/logics';
+
+export type SwapTokenParams = common.Declasifying<logics.zeroexv4.SwapTokenLogicParams>;
+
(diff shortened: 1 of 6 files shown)


packages/api/src/protocols/zeroex-v4/swap-token.ts
```diff
@@ -0,0 +1,25 @@
...
+import { getProtocolTokenList, quote } from 'src/api';
+import * as logics from '@protocolink/logics';
+
+export type SwapTokenParams = common.Declasifying<logics.zeroexv4.SwapTokenLogicParams>;
+
+export type SwapTokenFields = common.Declasifying<logics.zeroexv4.SwapTokenLogicFields>;
+
+export type SwapTokenLogic = Logic<SwapTokenFields>;
+
+export async function getSwapTokenTokenList(chainId: number): Promise<common.Token[]> {
```
zodahu: In the PR for protocolink-logics 0x, changes were made so that `paraswapv5` and `zeroexv4` use the common `TokenList` type `common.Token[]`.

To ensure consistency across all logics with a `TokenList` type, I suggest introducing a `logics.zeroexv4.SwapTokenList` type. This way, `getTokenList()` in both `paraswapv5` and `zeroexv4` can return their specific types. What are your thoughts?

Jeff-CCH: agreed. will do.

zodahu:
LGTM

PR #43, opened 2024-03-20
Add ZeroEx Test

Jeff-CCH:
- [x] add ZeroEx test and example
- [x] yarn changeset

### DIFF
--- a/packages/api/examples/zeroex-v4/swap-token.ts
+++ b/packages/api/examples/zeroex-v4/swap-token.ts
@@ -1,0 +1,82 @@
+import * as api from '@protocolink/api';
+
+// interface SwapTokenParams {
+//   input: {
+//     token: {
+//       chainId: number;
+//       address: string;
+//       decimals: number;
+//       symbol: string;
+//       name: string;
+//     };
+//     amount: string;
+//   };
+//   tokenOut: {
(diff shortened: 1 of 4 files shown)


packages/api/examples/zeroex-v4/swap-token.ts
```diff
@@ -0,0 +1,81 @@
...
+      amount: '10',
+    },
+    tokenOut,
+    slippage: 100,
+    apiKey,
+  });
+  swapTokenQuotation.apiKey = '';
+  console.log('swapTokenQuotation :>> ', JSON.stringify(swapTokenQuotation, null, 2));
+
+  const swapTokenLogic = await api.protocols.zeroexv4.newSwapTokenLogic(swapTokenQuotation);
```
zodahu: Given that `swapTokenQuotation` lacks an `apiKey`, it is anticipated that `newSwapTokenLogic` will generate an error.

Jeff-CCH: Actually, it won't since the `newSwapTokenLogic` only does a wrap for the params which means we don't need to reset the `apiKey` in `swapTokenLogic`.

zodahu:
LGTM

PR #44, opened 2024-03-28
feat(lending): support LST APY

LuPoYi:
- [x] support lst apy
- [x] fix unit test
- [x] yarn changeset
- [x] lst apy use dinngo aws bucket


packages/lending/src/adapter.ts
```diff
@@ -500,7 +500,7 @@ export class Adapter extends common.Web3Toolkit {
         if (flashLoanRepayLogic) output.logics.push(flashLoanRepayLogic);
 
         if (withdrawalToken.isNative && Number(withdrawal.amount) > 0) {
-          const wrapNativeLogic = await apisdk.protocols.utility.newWrappedNativeTokenLogic({
+          const wrapNativeLogic = apisdk.protocols.utility.newWrappedNativeTokenLogic({
```
Jeff-CCH, line 503: Can you help remove other `await` before `newWrappedNativeTokenLogic`?

packages/lending/src/protocol.portfolio.ts
```diff
@@ -135,6 +163,18 @@ export class Borrow implements AssetInfo {
     return formatPercentage(this.apy);
   }
 
+  get formattedLstAPY() {
+    return formatPercentage(this.lstApy);
+  }
+
+  get grossApy() {
+    return this.grossApys[0];
```
Jeff-CCH: What's the meaning of `grossApys[0]` and why do we get this value only ?

packages/lending/src/protocols/aave-v2/lending-protocol.ts
```diff
@@ -297,11 +299,17 @@ export class LendingProtocol extends Protocol {
         usageAsCollateralEnabled = usageAsCollateralEnabled && userBalance.usageAsCollateralEnabled;
       }
 
+      const lstApy = lstTokenAPYMap[token.address.toLowerCase()] || '0';
+      const grossApy =
+        lstApy === '0' ? reserveData.supplyAPY : BigNumberJS(reserveData.supplyAPY).plus(lstApy).toString();
```
Jeff-CCH: Since the interface of BIgNumberJS.plus() is plus(n: BigNumber.Value ...), I suggest transforming the `lstApy` from `string` to `BigNumberJS` to avoid any unexpected error (though the calculation is correct in the current implementation). What do you think?

packages/lending/src/protocols/aave-v2/lending-protocol.test.ts
```diff
@@ -620,7 +621,12 @@ describe('Test Aave V2 LendingProtocol', function () {
       it(`${common.toNetworkId(chainId)} market`, async function () {
         const protocol = new LendingProtocol(chainId);
         protocol.setBlockTag(blockTag);
-        const portfolio = await protocol.getPortfolio(account);
+        const _portfolio = await protocol.getPortfolio(account);
+        const portfolio = JSON.parse(JSON.stringify(_portfolio));
+
+        removePortfolioDynamicFields(expected);
+        removePortfolioDynamicFields(portfolio);
```
Jeff-CCH, line 654: Why do we remove the newly added fields?

packages/lending/src/protocols/aave-v3/lending-protocol.ts
```diff
@@ -295,11 +296,17 @@ export class LendingProtocol extends Protocol {
         ? userBalances.usageAsCollateralEnabled
         : reserveData.usageAsCollateralEnabled;
 
+      const lstApy = lstTokenAPYMap[token.address.toLowerCase()] || '0';
+
```
Jeff-CCH:
```suggestion
```

packages/lending/src/protocols/compound-v3/lending-protocol.ts
```diff
@@ -313,11 +320,17 @@ export class LendingProtocol extends Protocol {
       },
     ];
     for (const { token, borrowCollateralFactor, liquidateCollateralFactor, supplyCap, totalSupply } of assets) {
+      const apy = '0';
+      const lstApy = lstTokenAPYMap[token.address.toLowerCase()] || '0';
+      const grossApy = lstApy === '0' ? apy : BigNumberJS(apy).plus(lstApy).toString();
```
Jeff-CCH: Since apy = 0 in Ln.323, do we still need to do the `plus` operation here?

packages/lending/src/protocols/compound-v3/lending-protocol.ts
```diff
@@ -326,12 +339,15 @@ export class LendingProtocol extends Protocol {
       });
     }
 
+    const borrowGrossApy = lstApy === '0' ? borrowAPR : BigNumberJS(borrowAPR).minus(lstApy).toString();
```
Jeff-CCH: Do you think adding the condition check `lstApy === '0'` into the utility is a good way to keep the code clean?

packages/lending/src/protocols/morphoblue/lending-protocol.ts
```diff
@@ -135,25 +135,38 @@ export class LendingProtocol extends Protocol {
 
     const maxLtv = common.toBigUnit(lltv, 18);
 
+    const lstTokenAPYMap = await this.getLstTokenAPYMap(this.chainId);
+
+    const supplyApy = '0';
+    const supplyLstApy = lstTokenAPYMap[collateralToken.address.toLowerCase()] || '0';
+    const supplyGrossApy = supplyLstApy === '0' ? supplyApy : BigNumberJS(supplyApy).plus(supplyLstApy).toString();
```
Jeff-CCH: same as the `apy` comment

packages/lending/src/protocols/morphoblue/lending-protocol.ts
```diff
@@ -135,25 +135,38 @@ export class LendingProtocol extends Protocol {
 
     const maxLtv = common.toBigUnit(lltv, 18);
 
+    const lstTokenAPYMap = await this.getLstTokenAPYMap(this.chainId);
+
+    const supplyApy = '0';
```
Jeff-CCH, line 142: any reason why we don't use `apy` directly?

packages/lending/src/protocols/morphoblue/lending-protocol.ts
```diff
@@ -135,25 +135,38 @@ export class LendingProtocol extends Protocol {
...
+        lstApy: supplyLstApy,
+        grossApy: supplyGrossApy,
         usageAsCollateralEnabled: true,
         ltv: maxLtv,
         liquidationThreshold: maxLtv,
         totalSupply,
       },
     ];
 
+    const borrowLstApy = lstTokenAPYMap[loanToken.address.toLowerCase()] || '0';
```
Jeff-CCH: Do you think putting this line into utility is a good way to access the map?

packages/lending/src/protocol.portfolio.ts
LuPoYi: There are two types of APY for borrowing: `variableBorrowAPY` and `stableBorrowAPY` in aave-like protocol, so the grossApy is calculated in the same manner.
but the trush is that stable APY is almost unsupported, so we only use first one `variableBorrowAPY`

packages/lending/src/protocols/aave-v2/lending-protocol.test.ts
LuPoYi: In the test case, we set the `blockTag` to ensure the portfolio remains consistent within the specified block. However, lstApy is a real-time APY fetched from an external API, meaning the expected and portfolio values might not match.

packages/lending/src/protocols/morphoblue/lending-protocol.ts
LuPoYi: There's a `borrowApy` within the same scope, using `supplyApy` is fine, but sure we can rename to `apy` for better readability

packages/lending/src/protocols/aave-v2/lending-protocol.ts
LuPoYi: https://mikemcl.github.io/bignumber.js/#plus
```
BigNumber.Value
type Value = string | number | Instance;
```
I believe using a `string` as property is fine

packages/lending/src/protocols/aave-v3/lending-protocol.ts
LuPoYi: removed

packages/lending/src/protocols/compound-v3/lending-protocol.ts
LuPoYi: fixed

packages/lending/src/adapter.ts
LuPoYi: sure, there are two ones in api package, will remove them after this mr

packages/lending/src/protocols/compound-v3/lending-protocol.ts
LuPoYi: added `calcSupplyGrossApy` and `calcBorrowGrossApy` to make the code clean

packages/lending/src/protocols/morphoblue/lending-protocol.ts
LuPoYi: added `calcSupplyGrossApy` and `calcBorrowGrossApy` to make the code clean

packages/lending/src/protocols/morphoblue/lending-protocol.ts
LuPoYi: added the func `getLstApyFromMap` to access the map

packages/lending/src/protocols/aave-v2/lending-protocol.test.ts
Jeff-CCH: I see. Do we have any test to ensure the lst apy is good to go?

packages/lending/src/protocol.utils.ts
```diff
@@ -125,3 +125,27 @@ export const removePortfolioDynamicFields = (portfolio: any) => {
     });
   }
 };
+
+export const getLstApyFromMap = (address: string, lstTokenAPYMap: Record<string, string>) => {
```
Jeff-CCH: can we change `address` to `tokenAddress` to be clearer?

packages/lending/src/protocol.utils.ts
```diff
@@ -125,3 +125,27 @@ export const removePortfolioDynamicFields = (portfolio: any) => {
...
+  return lstTokenAPYMap[address.toLowerCase()] || '0';
+};
+
+export const calcSupplyGrossApy = (apy: string, lstApy: string) => {
+  if (lstApy === '0') {
+    return apy;
+  } else if (apy === '0') {
+    return lstApy;
+  } else {
+    return BigNumberJS(apy).plus(lstApy).toString();
```
Jeff-CCH: suggest using
`return lstApy === '0' ? apy : apy === '0' ? lstApy : BigNumberJS(apy).plus(lstApy).toString();`
to be cleaner

packages/lending/src/protocol.utils.ts
```diff
@@ -125,3 +125,27 @@ export const removePortfolioDynamicFields = (portfolio: any) => {
...
+  }
+};
+
+export const calcBorrowGrossApy = (apy: string, lstApy: string) => {
+  if (lstApy === '0') {
+    return apy;
+  } else if (apy === '0') {
+    return lstApy;
+  } else {
+    return BigNumberJS(apy).minus(lstApy).toString();
```
Jeff-CCH: same as `calcSupplyGrossApy`

Jeff-CCH: wondering if we need to make sure `apy > lstApy`?

packages/lending/src/protocols/morphoblue/lending-protocol.ts
Jeff-CCH: i see. In that case, `supplyApy` is easier to understand. thx.

packages/lending/src/protocols/compound-v3/lending-protocol.ts
```diff
@@ -321,8 +322,8 @@ export class LendingProtocol extends Protocol {
     ];
     for (const { token, borrowCollateralFactor, liquidateCollateralFactor, supplyCap, totalSupply } of assets) {
       const apy = '0';
```
Jeff-CCH, line 325: Do you think it's a good way that:
1. define a constant in config and use it to illustrate that compound v3 does not have any apy for supplying collaterals.
2. add a comment to explain why we give 0 here.

Which way do you prefer?

packages/lending/src/protocols/aave-v2/lending-protocol.test.ts
LuPoYi: The LST APY list link is built on multiple API requests, making it hard to verify if it's correct. However, if the link is invalid, we will ignore LST APY and set it to 0, and the gross APY will equal the original APY, I think that's not a big problem.

packages/lending/src/protocols/compound-v3/lending-protocol.ts
LuPoYi: 2, I think adding a comment is okay for that, I will add on morphoblue and compoundv3

packages/lending/src/protocol.utils.ts
LuPoYi: If the gross APY is positive, it means the borrower should pay interest. If the gross APY is negative, it means the borrower earns interest, so we don't need to keep it positive.

LuPoYi:
Fixed according to the suggestions and removed the borrow stable rate, merging them into a single string: `apy`, `grossApy`, `balance`

LuPoYi (commented):
Fixed according to the suggestions and removed the borrow stable rate, merging them into a single string: `apy`, `grossApy`, `balance`

packages/lending/src/protocol.utils.ts
LuPoYi: changed

packages/lending/src/protocol.utils.ts
LuPoYi: changed

packages/lending/src/protocol.utils.ts
LuPoYi: changed

zodahu:
LGTM

Jeff-CCH (approved):
LGTM

PR #45, opened 2024-04-03
feat(common): add logoURI prop to token interface

LuPoYi:
- [x] add logoURI(optional) prop to token interface
- [x] rename zksync USDC to USDC.e
- [x] yarn changeset

### DIFF
--- a/packages/common/src/tokens/token.ts
+++ b/packages/common/src/tokens/token.ts
@@ -8,6 +8,7 @@ export interface TokenObject {
   decimals: number;
   symbol: string;
   name: string;
+  logoURI?: string;
 }
 
 export class Token {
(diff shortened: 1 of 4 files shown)


packages/common/src/tokens/token.ts
```diff
@@ -16,8 +17,9 @@ export class Token {
   readonly decimals: number;
   readonly symbol: string;
   readonly name: string;
+  readonly logoURI: string | undefined;
```
zodahu:
```suggestion
  readonly logoUri?: string;
```

Simplify it and currently we use camelCase for naming, such as `chainId`, `protocolId` and `lstApy`.

packages/common/src/tokens/token.ts
zodahu, line 1: 1. In the `isTokenObject` function, if `logoURI` exists, we can verify that its type is a string.
2. Could we include a test for the newly added `logoURI` field in `tokentest.ts`?

LuPoYi: fixed!

packages/common/src/tokens/token.ts
LuPoYi: fixed

zodahu:
LGTM

PR #53, opened 2024-05-07
feat: add Protocolink Smart Account SDK

Jeff-CCH:
- [x] add Protocolink Smart Account SDK
- [x] test
- [x] yarn changeset


packages/smart-account/package.json
```diff
@@ -0,0 +1,37 @@
+{
+  "name": "@protocolink/smart-accounts",
```
zodahu: As this package name is @protocolink/smart-account`s`, please check all namings in this repo are plural. including the folder name.

packages/smart-account/src/abis/Executor.json
zodahu, line 1: Should the file name be `PortusExecutor`?

packages/smart-account/src/configs.ts
```diff
@@ -0,0 +1,60 @@
+import * as common from '@protocolink/common';
+
+export enum SmartAccountId {
+  PORTUS = '1',
```
zodahu:
```suggestion
  PORTUS = 'portus',
```
Prefer to give a readable key rather than serial number

packages/smart-account/src/index.ts
```diff
@@ -0,0 +1,17 @@
+import { Executor__factory } from './contracts';
+import { getConfig, getExecutor, getExecutorId } from './configs';
+
+export function getSmartIds(chainId: number) {
```
zodahu:
```suggestion
export function getSmartAccountIds(chainId: number) {
```
Prefer to keep the naming consistently

packages/smart-account/src/index.ts
```diff
@@ -0,0 +1,17 @@
...
+  return getExecutorId(chainId);
+}
+
+export function verifySmartId(chainId: number, id: string) {
```
zodahu:
```suggestion
export function isSupportedSmartAccountId(chainId: number, id: string) {
```
Prefer to keep the naming consistent with other npm in this repo such as `isSupportedChainId` and `isSupportedNetworkId`

packages/smart-account/src/configs.ts
```diff
@@ -0,0 +1,60 @@
...
+  {
+    chainId: common.ChainId.optimism,
+    executors: [
+      {
+        id: SmartAccountId.PORTUS,
+        address: '0xdD408b8eFb837EdeF8e6192Ed19f0dbEB7B79383',
+      },
+    ],
+  },
+];
```
zodahu:
```suggestion
export const configs: Config[] = [
  {
    chainId: common.ChainId.optimism,
    smartAccounts: [
      {
        id: SmartAccountId.PORTUS,
        executor: '0xdD408b8eFb837EdeF8e6192Ed19f0dbEB7B79383',
      },
    ],
  },
];
```

Looks like the object key is more like smart accounts as the `PORTUS` is smart account id rather than executor's id.

packages/smart-account/package.json
Jeff-CCH: I was thinking if we need the 's' and forget to remove this one. Do you prefer the 's' or not?

packages/smart-account/src/abis/Executor.json
Jeff-CCH: Yes. But it would be a little bit weird if we use a specific name (eg. PortusExecutor) to do the encoding for all smart accounts (because smart accounts share the same interface). That's the reason why I keep it as a general Executor (though it is generated from PortusExecutor).

zodahu: It's good that `Executor.json` is expected to serve as the common interface. Therefore, I think we should retain the functions that is common in this file. For example, function `isModuleType`is specific to the Protus wallet.

packages/smart-account/package.json
zodahu: I prefer `s` as this package is expected to aggregate many smart accounts, and plural in npm package name seems common used.

packages/smart-account/src/abis/Executor.json
Jeff-CCH: Is it a good idea that we only keep `executeFromAgent` in the abi json file?

zodahu: Sure, given that this is the only one function we used.

.changeset/ten-actors-learn.md
```diff
@@ -0,0 +1,5 @@
+---
+'@protocolink/smart-accounts': patch
+---
+
+The first version release for Protocolink Smart Account
```
zodahu:
```suggestion
The first version release for Protocolink Smart Accounts
```

packages/smart-accounts/README.md
```diff
@@ -0,0 +1,17 @@
+# @protocolink/smart-accounts
+
+The smart account sdk for Protocolink.
```
zodahu:
```suggestion
The smart accounts sdk for Protocolink.
```

packages/smart-accounts/package.json
```diff
@@ -0,0 +1,37 @@
+{
+  "name": "@protocolink/smart-accounts",
+  "version": "0.1.0",
+  "description": "Protocolink Smart Account SDK",
```
zodahu:
```suggestion
  "description": "Protocolink Smart Accounts SDK",
```

packages/smart-accounts/src/configs.ts
```diff
@@ -0,0 +1,60 @@
...
+    smartAccounts: [
+      {
+        id: SmartAccountId.PORTUS,
+        executor: '0xdD408b8eFb837EdeF8e6192Ed19f0dbEB7B79383',
+      },
+    ],
+  },
+];
+
+export const [supportedChainIds, configMap, executorMap, executorIdMap] = configs.reduce(
```
zodahu:
```suggestion
export const [supportedChainIds, configMap, smartAccountMap, smartAccountIdMap] = configs.reduce(
```

Please also update `export function getXXX` naming

packages/smart-accounts/src/index.ts
```diff
@@ -0,0 +1,17 @@
+import { Executor__factory } from './contracts';
+import { getConfig, getSmartAccount, getSmartAccountId } from './configs';
+
+export function getSmartAccountIds(chainId: number) {
+  return getSmartAccountId(chainId);
```
zodahu:
```suggestion
  return getSmartAccountIds(chainId);
```

zodahu:
LGTM

PR #58, opened 2024-05-15
feat(lending): add BNB Chain info to Radiant V2

Jeff-CCH:
- [x] add BNB Chain to Radiant V2
- [x] yarn changeset


packages/lending/src/adapter.config.ts
```diff
@@ -66,6 +67,12 @@ export const configs: Config[] = [
     secondaryStablecoin: gnosisTokens.USDC,
     primaryNonstablecoin: gnosisTokens.WETH,
   },
+  {
+    chainId: common.ChainId.bnb,
+    primaryStablecoin: bnbTokens.USDT,
+    secondaryStablecoin: bnbTokens.USDC,
+    primaryNonstablecoin: bnbTokens.BNB,
+  },
```
zodahu, line 75: Have you confirmed the setup conventions with the FE, and if the native token `BNB` can be used in this list?

packages/lending/src/protocols/radiant-v2/lending-protocol.test.ts
```diff
@@ -14,20 +15,13 @@ describe('Test Radiant V2 LendingProtocol', function () {
...
           supplies: [
             {
-              token: {
-                chainId: 1,
-                address: '0xdAC17F958D2ee523a2206206994597C13D831ec7',
-                decimals: 6,
-                symbol: 'USDT',
-                name: 'Tether USD',
-              },
+              token: mainnetTokens.USDT,
```
zodahu, line 24: It's cool to use the token mapping, however I guess the `expected` fields are copied and pasted from the log, so when we update a new `blockTag`, we don't need to manually change the token mapping. Do you think this change is better for maintenance?

packages/lending/src/protocols/radiant-v2/configs.ts
```diff
@@ -62,13 +62,62 @@ export const configs: Config[] = [
         rToken: mainnetTokens.rwstETH,
         used: { deposit: true, withdraw: true, borrow: true, repay: true, flashLoan: true },
       },
+      {
+        asset: mainnetTokens.sDAI,
+        rToken: mainnetTokens.rsDAI,
+        used: { deposit: true, withdraw: true, borrow: true, repay: true, flashLoan: true },
+      },
```
zodahu, line 69: We didn't add sDAI initially because the price on Paraswap V5 was very bad at that time. However, I just checked, and it seems Paraswap V5 now supports the routing of sDAI to DAI through mint/burn. Therefore, it's a good time to add sDAI!

packages/lending/src/adapter.config.ts
Jeff-CCH: Yes. I checked the tokens with @LuPoYi.

packages/lending/src/protocols/radiant-v2/lending-protocol.test.ts
Jeff-CCH: Yeah. I think it's better for maintenance because

1. It shrink the size of `lending-protocol.test.ts` and make it easier to read
2. I understand it is easier to copy paste from the log at the beginning, but when we need to add a new token into this file, it would need to manually key-in the data to avoid clause issue.
3. The other benefit will be checking the info in the token mapping again in case of any typo.

packages/lending/src/protocols/radiant-v2/configs.ts
Jeff-CCH: I also checked this with @LuPoYi and Dan and they agreed to add sDAI back.

packages/lending/src/protocols/radiant-v2/lending-protocol.test.ts
zodahu: OK. c.c. @LuPoYi this change.

packages/lending/src/adapter.config.ts
zodahu: OK, I created a ticket to track that Base's primaryNonstablecoin is native token.

packages/lending/src/protocols/radiant-v2/configs.ts
zodahu: OK, it seems they weren't aware of the unfavorable price, anyway let's add sDAI.

zodahu (approved):
LGTM

PR #59, opened 2024-05-27
fix(lending): config primaryNonstablecoin

LuPoYi:
- Set all primaryNonstablecoin tokens on each chain to the native token


zodahu (dismissed):
LGTM

zodahu (approved):
LGTM

PR #61, opened 2024-06-06
feat(lending): support blocktag for adapter getPortfolio

LuPoYi:
- [x] add optional `blocktag` props for adapter `getPortfolio` & `getPortfolios`
- [x] yarn changeset


packages/lending/src/adapter.ts
```diff
@@ -159,11 +159,17 @@ export class Adapter extends common.Web3Toolkit {
...
+   * @param {Object} [options] - Optional parameters.
+   * @param {string|number} [options.blockTag] - An optional block tag
    * @returns {Promise<Portfolio[]>} A promise that resolves to an array of Portfolio objects from all protocols.
    */
-  async getPortfolios(account: string): Promise<Portfolio[]> {
+  async getPortfolios(account: string, options?: { blockTag?: string | number }): Promise<Portfolio[]> {
     const portfolios = await Promise.all(
-      Object.values(this.protocolMap).map((protocol) => protocol.getPortfolios(account))
+      Object.values(this.protocolMap).map((protocol) => {
+        if (options?.blockTag) protocol.setBlockTag(options.blockTag);
```
weizard: Should we always set the block tag? If I want to retrieve the current block but have previously set a block tag, the system will use the prior block tag setting instead of fetching the latest block.

LuPoYi: Good point. It seems that when the client doesn't pass the block tag, we need to clear the block tag so that it defaults to the latest block.

PR #68, opened 2024-07-17
fix(test-helpers): get token information from common package

Jeff-CCH:
- [x] get token info from common package
- [x] test
- [x] yarn changeset


LuPoYi:
LGTM

PR #76, opened 2024-07-28
test: add Magic Sea test and example

Jeff-CCH:
- [x] add Magic Sea test and example
- [x] use token info from common package
- [x] yarn changeset


LuPoYi:
LGTM

PR #83, opened 2024-08-15
fix(common): update polygon zkevm token info

Jeff-CCH:
- [x] update polygon zkevm token info
- [x] yarn changeset


LuPoYi:
LGTM

PR #84, opened 2024-08-20
feat(lending): optimize morphoblue getPortfolios multicall

LuPoYi:
- [x] yarn changeset
- [x] test
- [x] update morphoblue naming
- [x] optimize morphoblue multicall


packages/lending/src/protocols/morphoblue/configs.ts
```diff
@@ -3,7 +3,7 @@ import * as logics from '@protocolink/logics';
 import { mainnetTokens } from './tokens';
 
 export const ID = 'morphoblue';
-export const DISPLAY_NAME = 'Morpho Blue';
+export const DISPLAY_NAME = 'Morpho';
```
Jeff-CCH, line 6: I thought the DISPLAY_NAME is the same as the folder name. If we would like to use Morpho, should we change the folder name too?

packages/lending/src/protocols/compound-v3/configs.ts
```diff
@@ -13,6 +13,11 @@ export interface AssetConfig {
   supplyCap: string;
 }
 
+export interface PriceFeedConfig {
```
Jeff-CCH, line 16: MarketConfig could `extends PriceFeedConfig`?

packages/lending/src/protocols/morphoblue/configs.ts
LuPoYi: `DISPLAY_NAME` is only used on the front end for display purposes. A similar case is `zeroex-v4`, where if we needed to set a display name for it, it would be `0x`, which doesn't necessarily match the folder name.

And I believe 'Morpho Blue' is still the brand name, and they just use 'Morpho' for the lending, so I don't think we need to change the folder name unless they decide to rebrand.

packages/lending/src/protocols/compound-v3/configs.ts
LuPoYi: Missed that, thanks

Jeff-CCH (approved):
LGTM

PR #92, opened 2026-09-05
fix(common): decode symbol and name independently in web3-toolkit

Tyagiquamar:
In getTokens (packages/common/src/web3-toolkit.ts), j was incremented within the combined try block for symbol and name decoding. If symbol decoding succeeded but name decoding threw (requiring bytes32 fallback), the catch block re-used the already-incremented j value, causing bytes32 decoding to read from the wrong index and shifting all subsequent token decodings in the multicall result.

This change decodes symbol and name independently with explicit indices symbolIndex and nameIndex, ensuring token attributes decode reliably even if token contracts use mixed standard string and bytes32 formats.

### DIFF
--- a/packages/common/src/web3-toolkit.ts
+++ b/packages/common/src/web3-toolkit.ts
@@ -72,9 +72,12 @@ export class Web3Toolkit {
         let name: string;
         try {
           [symbol] = this.erc20Iface.decodeFunctionResult('symbol', returnData[1]);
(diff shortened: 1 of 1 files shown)


Tyagiquamar:
Hi, friendly ping for review on this PR when you have a moment. Happy to address any feedback. Thanks!

Tyagiquamar:
Hi, just following up on this when you get a chance. The branch is up to date and checks are green. If it looks good from your side, it should be ready to merge. Happy to make any changes if needed. Thanks!

PR #93, opened 2026-09-12
fix(common): decode symbol and name independently in getTokens

Tyagiquamar:
## Description
- Separate symbol and name decoding in \Web3Toolkit.getTokens\ into independent \	ry/catch\ blocks.
- Previously, a single \	ry/catch\ wrapped both symbol and name decoding. If \symbol\ decoding succeeded as string and incremented \j\, but \
ame\ decoding failed (e.g. \ytes32\ name like MKR), the \catch\ block executed with an already-incremented \j\, reading \
eturnData[j]\ as symbol and corrupting array indices for all subsequent tokens.
- Aligns \getTokens\ decoding logic with \getToken\.

## Testing
- Extended \getTokens\ test suite in \packages/common/src/web3-toolkit.test.ts\ with \MKR\ bytes32 token test.

### DIFF
--- a/packages/common/src/web3-toolkit.test.ts
+++ b/packages/common/src/web3-toolkit.test.ts
@@ -57,10 +57,11 @@ describe('Web3Toolkit', function () {
           mainnetTokens.USDC.address,
           mainnetTokens.ETH.address,
           mainnetTokens.WETH.address,
+          mainnetTokens.MKR.address,
           ELASTIC_ADDRESS,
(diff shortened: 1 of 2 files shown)


Tyagiquamar:
Hi, just following up on this when you get a chance. The branch is up to date and checks are green. If it looks good from your side, it should be ready to merge. Happy to make any changes if needed. Thanks!

